PR #18, opened 2024-07-26
Disable updater in the admin settings

jmarmstrong1207:
This one was a pretty simple change. Obviously because it's a docker image, the update button is unnecessary and can cause confusions and breakage. 

What I changed was the constants.py file in cps/constants.py in calibre-web. admin.html checks for this constant for whether it should show or not, so I set it to false.

I have not tested this, so be sure to do so. but feel like it should work given that calibre-web has already implemented logic to disable the button

References:
 - [If statement for disabling updater button](https://github.com/janeczku/calibre-web/blob/f8adcbb69138a625d478b42808920bdbd3e4a7c6/cps/templates/admin.html#L237) 
 - [Where the constant is set in admin.html (imported from constants.py)](https://github.com/janeczku/calibre-web/blob/f8adcbb69138a625d478b42808920bdbd3e4a7c6/cps/admin.py#L64)
 - [Where the constant is set globally](https://github.com/janeczku/calibre-web/blob/f8adcbb69138a625d478b42808920bdbd3e4a7c6/cps/constants.py#L33)

### DIFF
--- a/calibre-web-automator/constants.py
+++ b/calibre-web-automator/constants.py
@@ -1,0 +1,199 @@
+# -*- coding: utf-8 -*-
+
+#   This file is part of the Calibre-Web (https://github.com/janeczku/calibre-web)
+#     Copyright (C) 2019 OzzieIsaacs, pwr
+#
+#   This program is free software: you can redistribute it and/or modify
+#   it under the terms of the GNU General Public License as published by
+#   the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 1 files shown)


crocodilestick:
This is a great start but I think we can use what's here to also display what CWA version users are on as well as to notify users of the availability of new CWA updates

jmarmstrong1207:
Done. Changed it to 1.2.2 to prepare for the next update

PR #25, opened 2024-07-27
Fixed some typos and chown commands

JPDVM2014:
I fixed some small user facing typos. 

I also removed what looks like a redundant chown command, and I changed another one to be consistent with the rest. I noticed some permission issues because the container is running as abc:users, but it is creating book folders using abc:1000.

I'm no programmer, so I apologize if there is a reason to use 1000 that I missed somewhere.

### DIFF
--- a/calibre-web-automator/check-cwa-install.sh
+++ b/calibre-web-automator/check-cwa-install.sh
@@ -35,7 +35,7 @@ fi
 echo ""
 
 if $cs && $bs && $mc; then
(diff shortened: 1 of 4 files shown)


jmarmstrong1207:
I think ideally, the app should have an env variable that can be changed in the docker-compose, then any references with `chown` will reference that @crocodilestick what do you think?

crocodilestick:
Hey @JPDVM2014 , there are internal reasons for some chown commands being `abc:1000` and some being `abc:users`, these are not typos or inconsistencies, the typos will be corrected but this branch will not be merged. Thanks for your help!

JPDVM2014:
I figured, thanks!

PR #38, opened 2024-07-30
Remove sleep 10s in new-book-detector

jmarmstrong1207:
This doesn't seem to be important anymore and also makes importing extremely slow for bulk downloads. Maybe add it as a small release 1.2.2-1?


crocodilestick:
Change applied to version 1.3.0

Issue #70, opened 2024-09-06
[bug] Cannot access the Web interface if ratelimited by github.


Buco7854:
**Describe the bug**
Sometimes I get this error message. It is caused by the fact that i get rate-limited by github api. A simple fix would be to check if the 'tag_name' is present and if not just act as if it was the same version.

```
Traceback (most recent call last):
File "/lsiopy/lib/python3.10/site-packages/flask/app.py", line 2190, in wsgi_app
response = self.full_dispatch_request()
File "/lsiopy/lib/python3.10/site-packages/flask/app.py", line 1486, in full_dispatch_request
rv = self.handle_user_exception(e)
File "/lsiopy/lib/python3.10/site-packages/flask/app.py", line 1484, in full_dispatch_request
rv = self.dispatch_request()
File "/lsiopy/lib/python3.10/site-packages/flask/app.py", line 1469, in dispatch_request
return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
File "/app/calibre-web/cps/usermanagement.py", line 35, in decorated_view
return login_required(func)(*args, **kwargs)
File "/lsiopy/lib/python3.10/site-packages/flask_login/utils.py", line 290, in decorated_view
return current_app.ensure_sync(func)(*args, **kwargs)
File "/app/calibre-web/cps/web.py", line 806, in index
return render_books_list("newest", sort_param, 1, page)
File "/app/calibre-web/cps/web.py", line 417, in render_books_list
return render_title_template('index.html', random=random, entries=entries, pagination=pagination,
File "/app/calibre-web/cps/render_template.py", line 153, in render_title_template
cwa_update_notification()
File "/app/calibre-web/cps/render_template.py", line 141, in cwa_update_notification
update_available, current_verision, tag_name = cwa_update_available()
File "/app/calibre-web/cps/render_template.py", line 119, in cwa_update_available
tag_name = response.json()['tag_name']
KeyError: 'tag_name'
```

**To Reproduce**
Go to any page
**Expected behavior**
Display the page as normal

**Configuration(please complete the following information):**
 - OS: Debian
 - Version 2.0.1

**Additional context**
Cannot describe how to reproduce clearly, it seems to happen randomly.

crocodilestick:
@Buco7854 closed thanks to the merge of your PR fixing the issue!

PR #71, opened 2024-09-06
fix: error when github does not respond with expected data.

Buco7854:
In case of a ratelimit github does not answer with the expected `tag_name` resulting in an error and therefore making the UI unusable.

I propose to make it so that if the expected `tag_name` key is not present replace it by the current version so that it does not result in an error and does not trigger update notification.

### DIFF
--- a/root/app/calibre-web/cps/render_template.py
+++ b/root/app/calibre-web/cps/render_template.py
@@ -116,7 +116,7 @@ def cwa_update_available() -> tuple[bool, str, str]:
     with open("/app/CWA_RELEASE", 'r') as f:
         current_version = f.read().strip()
     response = requests.get("https://api.github.com/repos/crocodilestick/calibre-web-automated/releases/latest")
-    tag_name = response.json()['tag_name']
+    tag_name = response.json().get('tag_name', current_version)
     return (tag_name != current_version), current_version, tag_name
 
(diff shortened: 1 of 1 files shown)


Buco7854:
Should fix #70

Buco7854:
I think this should be seen more as an temporary mesure to impelment a check that would not trigger rate limit that easily. Making a request to github API each time someone access a page in the web ui make it so that you get ratelimited pretty fast, and without even talking about the fact that it blocks the ui completly this may block other services using the api in the network. Maybe a button to check the version in the admin panel would be better. Or it could be completly removed, since its checking against github there is already a way to be alerted when a new released is out so I dont think this is really needed.

crocodilestick (approved):

crocodilestick:
> I think this should be seen more as an temporary mesure to implement a check that would not trigger rate limit that easily. Making a request to github API each time someone access a page in the web ui make it so that you get ratelimited pretty fast, and without even talking about the fact that it blocks the ui completly this may block other services using the api in the network. Maybe a button to check the version in the admin panel would be better. Or it could be completly removed. Since it's checking against github there is already a way to be alerted when a new released is out by github itself so I dont think this is really needed.

The function has also been changed to only send requests once per day to check for available updates

Buco7854:
Perfect, thanks!

PR #101, opened 2024-09-29
convert-library - don't copy original book when failed to convert the file | import converted book first before rm existing book | move failed import to processed_books/failed

jmarmstrong1207:
i have a feeling convert-library in 2.1.0 shouldn't copy failed conversions to processed_books/failed, but simply only log that book file conversion error. Makes sense for ingest folder to move failed adds to that folder, but for convert-library, it'd be wasting storage space because its just duplicating it 

convert-library also seems to have a logical bug where if importing converted book to library fails, the user is left with a calibre library where the original book is completely deleted and the converted book is just stuck in the tmp folder

For this, we could have it import converted book first, then remove the original book in the library after. So if it failed to import converted book, they at least have the book still in the library. 

And when it fails to import converted book, we can move this to processed_books/failed which it doesn't currently do, so they at least have the converted books on hand to manually import later possibly with cwa-ingest

### DIFF
--- a/scripts/convert-library.py
+++ b/scripts/convert-library.py
@@ -101,16 +101,6 @@ class LibraryConverter:
                 print_and_log(f"[convert-library]: Conversion of {os.path.basename(file)} successful! Removing old version from library...")
             except subprocess.CalledProcessError as e:
                 print_and_log(f"[convert-library]: Conversion of {os.path.basename(file)} was unsuccessful. See the following error:\n{e}")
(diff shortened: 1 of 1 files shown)


B0G0311:
Am anxiously waiting for this single fix before starting to use this app again. I am really hoping this is close to being released.

jtha:
I wish I read this issue before following the post-install instructions to run the convert-library command... Everything not a .epub is gone...

PR #151, opened 2024-12-20
[bug] kindle_epub_fixer used as function name and as boolean

Aymendje:
When `kindle_epub_fixer` is `true`, we try to call the function in the same class `kindle_epub_fixer`, but since it got overwrite by init to be a bool, we fail with 

```
[ingest-processor]: No conversion needed for ebook.epub, importing now...
Traceback (most recent call last):
  File "/app/calibre-web-automated/scripts/ingest_processor.py", line 257, in <module>
    main()
  File "/app/calibre-web-automated/scripts/ingest_processor.py", line 229, in main
    nbp.add_book_to_library(filepath)
  File "/app/calibre-web-automated/scripts/ingest_processor.py", line 167, in add_book_to_library
    self.kindle_epub_fixer(book_path)
```

This PR renames the bool to avoid the issue


crocodilestick:
Thank you!

PR #160, opened 2025-01-03
add ibdb.dev as a metadata provider

chad3814:
https://ibdb.dev/ is a book metadata service I set up, backed by https://isbndb.com/

### DIFF
--- a/root/app/calibre-web/cps/metadata_provider/ibdb.py
+++ b/root/app/calibre-web/cps/metadata_provider/ibdb.py
@@ -1,0 +1,117 @@
+# -*- coding: utf-8 -*-
+
+#  This file is part of the Calibre-Web (https://github.com/janeczku/calibre-web)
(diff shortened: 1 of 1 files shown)


crocodilestick:
That's amazing! Great work! Only one thing, if you're hosting https://ibdb.dev/, what version of the API is it running on? The free keys only allow 1 request per second and so if  all of CWA's users start using this provider in the next version, there will be too many requests for it to handle no?

chad3814:
so I have the basic account, which is 5k requests a day and 1/sec. I cache results, so if multiple people have the same filename, it will only make 1 request to ISBNDb. If it becomes a problem, I can probably do some better searching/culling on the cached data. Also I can proactively fetch a list of books on publication date, so I can fill in a backlog and forelog, and then requery each day, and then do the searching on the cached data, and kick off a live search if there are no or only a few results.

PR #232, opened 2025-02-03
Update library refresh subprocess.run to subprocess.Popen to make non-blocking

robajr:
subprocess.run is a blocking system call, subprocess.Popen is non-blocking. Errors in import freeze / take down CWA. Switching to Popen makes this asynchronous.


robajr:
While this makes the process unblocking, the results need to be handled asynchronously. Will make a more robust changeset via asyncio (or other)

PR #270, opened 2025-02-18
Update README.md

Turmaxx:
updated port number on the quick start guide section of the README.md to match the port number provided by default.


gelbphoenix:
Fixes: #309

PR #284, opened 2025-02-28
feat: allow retaining original file after conversion

### LINKED ISSUE #269: [Feature Request] Add option to keep original format after conversion during auto-ingestion
**Is your feature request related to a problem? Please describe.**
Currently there seem to not be an option to keep the original format when converting to e.g. epub during auto-ingestion.
The original file simply is replaced by the new format.

**Describe the solution you'd like**
An option in the Settings to keep the original file/format after the conversion and auto-ingestion.

**Describe alternatives you've considered**
By disabling conversion of specific formats this can be avoided, but at the same time this also means that no conversion will take place at all for those formats which is most often not the wanted outcome.

**Additional context**
None

jyyyeung:
Added a new setting to allow retaining original book format in auto-ingestion. Formats enabled in that option will also be added to the library even if it has been converted to the target format. 

1. The target format will automatically be enabled. 
2. The format will be ignored if it has been enabled in the ignore ingest settings. 

image

Fixes #269

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -12,3 +12,4 @@ Dockerfile
 build
 .history
 Dockerfile_calibre_not_included
+temp
--- a/.gitignore
+++ b/.gitignore
@@ -182,3 +182,11 @@ cython_debug/
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
+
+temp/
+library/
+config/
+ingest/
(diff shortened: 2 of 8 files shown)


thechitowncubs:
Thank you, this is what I need.

crocodilestick:
Would love to merge but could you work on the conflicts? Please also clean up the PR, it's full of files that have changes that aren't related to the PR at all for silly things like changing all single quotes to double quotes in an unrelated file or changes to your ignore files. The functionality looks great though and I would love to merge ASAP

PR #308, opened 2025-03-10
Use replace to get the suffix only (no .)

qliratu:
Remove the . (dot) from the check, since ingest_ignored_formats gives a list whitout . (dots)


qliratu:
Feature is not working now, plain suffix gives a list with .'s

crocodilestick:
Thank you <3

PR #317, opened 2025-03-20
Always ignore .crdownload during ingestion

Aymendje:
This change updates the ingestion process to ignore files with the `.crdownload` extension.
Since browsers like Chrome use `.crdownload` for files in progress, this prevents partially downloaded files from being processed, which could lead to issues.


Aymendje:
We might want to add `.part` for Firefox support, and `.download` for Safari ?

gelbphoenix:
> We might want to add `.part` for Firefox support, and `.download` for Safari ?

Yes, we should better add all major browser with that and not only Chrome and Chromium based ones.

Aymendje:
Good idea, I updated it
@crocodilestick  please let me know if you want any changes !

crocodilestick:
Great job @Aymendje! Merged :)

PR #329, opened 2025-03-27
Fix for older kernels (E.G. synology) - Remove the ABI tag from the extracted libQt6* files

FennyFatal:
- Add `binutils` to install `strip` for calibre-included Dockerfile.
- strip `libQt6*.so` files of the `ABI` tag so that they can work with older kernels (harmless for newer kernels).
  - These libraries appear to still contain fallbacks for any missing syscalls that calibre might use.
- add .gitattributes to enforce LF checkout on `.sh` files (useful for those who build on windows)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.sh text eol=lf
--- a/Dockerfile
+++ b/Dockerfile
@@ -157,7 +157,8 @@ RUN \
     libxdamage1 \
     libgl1 \
     libglx-mesa0 \
-    xz-utils && \
+    xz-utils \
+    binutils && \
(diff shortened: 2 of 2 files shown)


crocodilestick:
Have you tested this on a system with an older kernal? This should let devices running old kernels run the latest version of Calibre no? Then the Version CWA uses wouldn't have to be locked to 7.16.0?

FennyFatal:
Yes, I have it working with 8.4 on an older NAS right now.

eLindros:
I use this patch successfully on my Synology DS916+. Without the patch calibre does not cleanly run in the container.

PR #361, opened 2025-04-26
Rootless Container and Read-Only Filesystem

jameswynn:
I don't run containers that require root in my cluster, so I forked and refactored the project such that it doesn't run as root and has a read-only filesystem. The big changes I made come down to:

- Move to standard ubuntu:22.04 as base image (no LSIO)
- Removed all chown commands
- Calling calibre_postinstall in Dockerfile instead of on startup
- Refactored s6 services to normal scripts in the CWA scripts directory
  - New entrypoint.sh that kicks off all the "services"
  - All RW data is under /config
    - dirs.json - copied on startup from /app/calibre-web-automated/dirs.json
    - metadata_change_logs
    - cwa_update_notice
    - metadata_temp


geekifier:
Just wanted to say thank you for your work on this! I don't see this feature mentioned in the [high priority](https://github.com/crocodilestick/Calibre-Web-Automated?tab=readme-ov-file#high-priority-) section of the README, but IMHO this refactor should be high up on the list.

FennyFatal:
I have some concerns that there are some people using the LS.io docker mods and init script runners which may not work well with an immutable filesystem.

Either way, this probably needs a rebase and further discussion, given recent merges.

distante:
Following! I hope this PR gets merged and released soon.

Issue #219, opened 2025-01-31
[bug] CWA library fails to find files that it's trying to convert


Waggleworx:
**Describe the bug**
My large library fails kobo syncs sometimes if files aren't already converted to kepub.  I decided to try using CWA Library Convertor after temporarily changing the target format from epub to kepub.  After completion, I did NOT find new kepub formats for all of my books

**To Reproduce**
1. Ingest epub books with epub as default
2. change target format to kepub
3. run convertor
4. check books that didn't have kepub

**Expected behavior**
I'd expect kepub for all books to be available

**Screenshots**
CWA log contains a lot of this for every file.  It looks like the path to move the starting file might not be constructed correctly since it can't find the file

```
[convert-library]: ERROR - The following error occurred when trying to copy Headhunted to Another World - From Salarym - Benigashira.epub to /config/processed_books/converted:
[Errno 2] No such file or directory: 'Headhunted to Another World - From Salarym - Benigashira.epub'
[convert-library]: (120/251) Intermediate conversion of Headhunted to Another World - From Salarym - Benigashira.epub to epub from .epub successful, now converting to kepub...
[convert-library]: ERROR - The following error occurred when trying to copy Headhunted to Another World - From Salarym - Benigashira.epub to /config/processed_books/converted:
[Errno 2] No such file or directory: 'Headhunted to Another World - From Salarym - Benigashira.epub'
[convert-library]: ERROR - The following error occurred when trying to copy /config/.cwa_conversion_tmp/Headhunted to Another World - From Salarym - Benigashira.kepub to /config/processed_books/imported:
[Errno 2] No such file or directory: '/config/.cwa_conversion_tmp/Headhunted to Another World - From Salarym - Benigashira.kepub'
[convert-library]: (120/251) Import of Headhunted to Another World - From Salarym - Benigashira.kepub successfully completed!
[convert-library]: (120/251) Successfully set ownership of new files in /calibre-library/ to abc:abc.
[convert-library]: (121/251) Converting Headhunted to Another World - From Salarym - Benigashira.epub from .epub format to kepub format...
```

**Configuration(please complete the following information):**
 - OS: Docker container
 - Hardware: Synology NAS
 - Version 3.0.1

**Additional context**
n/a

avephill:
experiencing this too on version 3.0.1 on docker container

crocodilestick:
Fixed as of V3.0.2

ajoliveau:
[...]

PR #371, opened 2025-05-09
Fix: Kepub conversion not being passed the full path of the source file.

Dee76:
When converting to **.kepub** only the filename of the source was being passed, not the full path. This resulted in an error message like: `[Errno 2] No such file or directory: '<example ebook>.epub'`.

The `convert_library.py` script was updated to pass the full path.

This fixes bug #219.


crocodilestick:
Thank you for catching this! Not being a Kobo user it's things like this I tend to miss 😅

PR #377, opened 2025-05-15
Hardcover updates

demitrix:
Fixes the following
Hardcover sync being required to enter token
Global Hardcover token
All meta items checked


crocodilestick:
@demitrix would it be possible for you to write a small thing for the new wiki explaining how people use your hardcover sync integration? I'm getting a lot of questions from users like the one below but I can't actually answer them as I've never had a Kobo to use your feature myself 😅

> How can I sync my Kobo progress to CWA? Is it separate from the kobo sync token I have setup to send books from shelves? Do I need to use KOReader or Koboutilities to sync progress?

PR #387, opened 2025-05-26
KOReader sync

demitrix:
Here is the initial work for a KO Reader sync that I started. This was more a proof of concept than anything else. Creating a PR to pass the buck. If someone wants to take these ideas and run with them.


FennyFatal:
This looks awesome.

Thinking about it:
- We probably would need to decouple the hash from the book records at least for tracking reading progress.
- Any arbitrary hash should allow for reading progress to be saved or loaded for a given user, even if that book is not in the library.
- On an action that would load the state for the existing reading tracking in the ui the hash can be populated (if it doesn't already exist) and at that time used for a lookup.
- In order to reduce the amount of initial hashing, hashes could be generated for books with existing reading progress instead of the whole collection, and then scheduled for the whole collection as a background task.
- Previous hashes should be kept for a given book id so that cover or metadata updates don't break sync state.

sirwolfgang:
Building on what @FennyFatal is saying, I think the roadmap for this feature looks like:
1. Replacing [koreader-sync-server](https://github.com/koreader/koreader-sync-server)
Tracks and syncs everything, but doesn't yet connect to CWA functionality; means one less container to deploy while solving authentication with CWA's profiles. This would automatically work with all documents and start collecting the data.

2. Pulling the data
This would include building out the pivot data to connect the files with digests that can be used to look up the progress. Then, on update, the process can be pushed to CWA's version of the tracking used in the UI and then eventually pushed to hardcover. I would avoid making the syncer and CWA's data the same because of a one-to-many relationship with file versions (and maybe editions, looking at hardcover).

Design Thoughts:
For the multiple versions question, koreader sync server offloads this to the UI of the reader to prompt (or use configurations) how to move forward. The result of this, after usage, is then back synced to the server. I suggest trusting that process, and just feeding it to CWA instead of coming up with a new UX that is outside of the usage loop. 

The hashing that KOReader uses is a pretty cheap MD5, so I think it's fine to generate those on import of the file if CWA isn't already. Having a file checksum like that, along with the tracking record, is good practice to detect corruption anyway. One flaw is that KOReader uses MD5s, which are pretty collidable, while rare, given that monkeys typing on keyboards is infinite, I would not be surprised if a collision does happen at some point; but that's an upstream problem.

mwinters0:
@crocodilestick I think this is now delivered, yes?

PR #408, opened 2025-06-27
Magic qrcode

coissac:
I developed a patch on Calibre-Web to add a QR Code to the magic-link login page.
The patch changes two files of the Calibre-Web source code:

- [`cps/remotelogin.py`](https://github.com/janeczku/calibre-web/blob/master/cps/remotelogin.py)
- [`cps/templates/remote_login.html`](https://github.com/janeczku/calibre-web/blob/master/cps/templates/remote_login.html)

The code added uses the python `qrcode` library

It allows an easier login by flashing the QRCode rather than copying the remote login URL when connecting using this feature

Have also added a `WORKDIR` directive at the end of your Dockerfile to insure the good running directory

All the best,

Eric

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 certifi==2024.7.4
 charset-normalizer==3.3.2
 idna==3.7
-requests==2.32.3
+requests==2.32.4
 tabulate==0.9.0
-urllib3==2.2.2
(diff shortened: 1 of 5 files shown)


crocodilestick:
Love it! Amazing work Eric, I'll work on merging ASAP :)

Issue #334, opened 2025-04-08
[Feature Request] Automatic Injest should merge books with same base name


doxxx:
**Is your feature request related to a problem? Please describe.**
When importing books/manga from something like HumbleBundle where multiple file formats are provided for each book, each format is imported as a separate book by CWA's automatic injest.

**Describe the solution you'd like**
I would like books with the same base name but different extensions to be injested as a single book. For example, `some_book_vol1.epub` and `some_book_vol1.pdf` should be imported as a single book with multiple formats, using the metadata from the epub.

**Describe alternatives you've considered**
The only alternative right now is to manually merge the books in CWA after they've been injested, which is extremely tedious for something like a manga HumbleBundle which could include 100 volumes (e.g. Humble Manga Bundle - Spring Shonen Special by Kodansha).

doxxx:
Alternatively, when injesting subfolders of the injest folder, CWA could treat every file in a particular subfolder as being different formats of the same book, like the calibre desktop import is able to do.

doxxx:
It looks like calibre's automatic book merging is being disabled when adding each book to the library: 

https://github.com/crocodilestick/Calibre-Web-Automated/blob/19960a926eaef4461d45a4634cd90ca6ba8f0ded/scripts/ingest_processor.py#L185

The commit where this was added doesn't document why this option was added to the `calibredb add` command. Maybe this could be made optional?

toeachbyeach:
> It looks like calibre's automatic book merging is being disabled when adding each book to the library:
> 
> [Calibre-Web-Automated/scripts/ingest_processor.py](https://github.com/crocodilestick/Calibre-Web-Automated/blob/19960a926eaef4461d45a4634cd90ca6ba8f0ded/scripts/ingest_processor.py#L185)
> 
> Line 185 in [19960a9](/crocodilestick/Calibre-Web-Automated/commit/19960a926eaef4461d45a4634cd90ca6ba8f0ded)
> 
>  subprocess.run(["calibredb", "add", book_path, "--automerge", "new_record", f"--library-path={self.library_dir}"], check=True) 
> The commit where this was added doesn't document why this option was added to the `calibredb add` command. Maybe this could be made optional?

Could this also help stop duplicates? This is the main problem I'm having.

avidenic:
[...]

Issue #208, opened 2025-01-30
problems importing duplicate books


frrad:
I would like to be able to import a book which is a more recent version but otherwise a duplicate of a book already in my library. 

Currently, import fails with `The following books were not added as they already exist in the database (see --duplicates option or --automerge option)`. 

1. Ideally for my use case I would like to call add with `overwrite`, so I guess feature requests = expose some knob where I can configure the argument here https://github.com/crocodilestick/Calibre-Web-Automated/blob/4951b2928931843892bc7d35251fb67224d238d0/scripts/ingest_processor.py#L179
2. The current release does not seem to include https://github.com/crocodilestick/Calibre-Web-Automated/commit/a8eb49a96afae20054db8ab436cce2fb4e66dda5 so the current behavior just deletes the file altogether and I'm left having to delete the original book before being able to import the new one. Can you please cut a new release including this commit? I am using the recommended docker compose configuration.

frrad:
(just edited to reflect the fact that the 3.0.1 release addressed part of what was in this issue originally)

burkesquires:
I would like to second this. I do not see a way to upvote

crocodilestick:
Added in V3.1.0

PR #417, opened 2025-07-09
Add configurable setting for ingest automerge parameter 

have-a-boy:
This is an attempt to address #340 #334 #208 as well as my own desire to overwrite duplicate books on import by adding a new setting that matches the three possible values of [calibredb's `--automerge` parameter](https://manual.calibre-ebook.com/generated/en/calibredb.html#cmdoption-calibredb-add-automerge). This replaces the hardcoded `new_record` value in [`add_book_to_library`](scripts/ingest_processor.py). The default value matches the previous behaviour.

I have done my best to keep this fairly minimal and to match any existing patterns where I was able to identify them. Please let me know your thoughts on a feature like this in general as well as this PR.


crocodilestick:
This honestly is a great idea, no notes. Thank you!

Issue #427, opened 2025-07-18
[bug] Error editing book: Object of type datetime is not JSON serializable


alva-seal:
**Describe the bug**
If i try to edit the metadata of any book, testet multiple, and save I get the error:

`Error editing book: Object of type datetime is not JSON serializable`

**To Reproduce**
1. Go to the page of any Book
2. Click on edit metadata
3. Scroll down to save
4. The error will be on the bottom of the page and data is not saved

**Expected behavior**
That the changed metadata is stred

**Screenshots**
Error:

![Image](https://github.com/user-attachments/assets/7f0a4fd8-bdcd-402b-baf7-f6b0ed73e231)

**Configuration(please complete the following information):**
 - OS: [Debian

V3.0.4 mit Docker

**Additional context**

PulsarFTW:
Is your calibre library on a different device the the calibre web automated server? I.E do you access it over a network? Or have you accessed your calibre library from another device over the network?

I think this type of error can occur when the database is corrupted from network access. The calibre database is unstable when access over a network.

alva-seal:
Hello,

I do not access the data base via network. 
I only use it in docker on the server but not outside the server in any way. And on the server it is mounted as a volume to the docker container.

Interestingly if i shut down calibre web automated and instead use the original calibre web image, i can store the changes. But I think that’s not writing to the database.

I also tried and checked the database with an calibre image and did no errors in the database.

alva-seal:
Here is also the relevant part of the log
`[metadata-change-detector] New file detected: 20250728075340-11239.json
Traceback (most recent call last):
  File "/app/calibre-web-automated/scripts/cover_enforcer.py", line 346, in <module>
    main()
  File "/app/calibre-web-automated/scripts/cover_enforcer.py", line 324, in main
    log_info = enforcer.read_log()
  File "/app/calibre-web-automated/scripts/cover_enforcer.py", line 126, in read_log
    log_info = json.load(f)
  File "/usr/lib/python3.10/json/__init__.py", line 293, in load
    return loads(fp.read(),
  File "/usr/lib/python3.10/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.10/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.10/json/decoder.py", line 355, in raw_decode
[...]

PR #445, opened 2025-08-01
Added a fix for #427

alva-seal:
custom columns with dates where not converted to a string before the json converter and caused errors.

Before it needs to be date time for saving it in the database
Error was reported in #427

Is probably not a common problem that a user has a custom column using dates, but i have the use case so I tried to fix it and it seems to work for me with this small fix.

Feel free to change it


crocodilestick:
Thanks Alva! Merged ❤️

PR #471, opened 2025-08-03
Add scalable theme system with Swan theme and custom CSS overrides

jmswangit:
This is probably not even a little bit similar to what you have in mind and also probably not super helpful, but I wanted to spend some time today working in github on something and so I made a theming system for CWA. It's not perfect, by any means, and may not even be worth your time, but I figured I'd submit the pull request on the very tiny off-chance it makes your life easier in some way. If not, feel free to reject and ignore it. I appreciate your patience in advance haha

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -45,6 +45,7 @@ from sqlalchemy.sql.expression import func, or_, text
 from . import constants, logger, helper, services, cli_param
 from . import db, calibre_db, ub, web_server, config, updater_thread, gdriveutils, \
     kobo_sync_status, schedule
+from .theme_manager import theme_manager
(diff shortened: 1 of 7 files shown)


crocodilestick:
Hey, I'm swamped ATM but I'll get to this ASAP <3 Could you add some screenshots here in the meantime to show it off and explain things?

crocodilestick:
Why did you close this?

jmswangit:
Truthfully, I felt embarrassed by it. It was not very good and not very scalable and I felt like I would only be adding work to you in the long run.

crocodilestick:
No! No way, even if you're not happy with the CSS the theme manager looks like a great idea! It's on my roadmap though to make themes user scope instead of server wide and something like that is a step in the right direction

PR #488, opened 2025-08-06
Update kosync_plugin.html to removed iOS reference

alva-seal:
I removed the reference to iOS as koreader is not availible there (and no reader kompatible to kosync to my knowledge)


alva-seal:
Don’t know why but it put my second change for translations  in here too, sorry I’m not familiar with github PRs

crocodilestick:
Thanks Alva!

PR #498, opened 2025-08-11
Add support for uploading, ingesting, and converting .acsm and include additional plugin documentation

natabat:
Added acsm to the various file format lists (and the mimetypes list) so CWA accepts them for converts, ingestion, and uploads. Also, I found that calibre's customize.py.json also needs to be copied to get a fresh install working with plugins, so I added notes to the readme and docker-compose files.


crocodilestick:
Awesome, thank you @natabat <3

PR #508, opened 2025-08-11
Copy uploaded files to ingest directory instead of processing them

natabat:
On upload, copy the file(s) to the ingest directory instead of processing them manually.

Note: this is kind of a hack, I would not accept this PR as part of my day job. The better way would be to refactor the ingestion code so it can be called directly on upload, but I don't have a real dev environment setup for this and that seemed too hard over putty and nano.


natabat:
Revisiting this to (hopefully) do it the right way.

PR #518, opened 2025-08-13
Ingest on upload

natabat:
Copy files to the ingest directory when they're uploaded through the web interface. Add a task for each file to indicate to the user that it is working.

Note: I tried implementing calling the NewBookProcessor directly, but it hung the upload dialog until everything was processed, which seemed undesirable -- it would really need to be added to Calibre Web's task queue, and that was a huge refactor. Even though this is kind of hacky, it ended up being the simplest way with the least possibility of breaking stuff.


crocodilestick:
This is great work! I've added abit to it to avoid a few potential issues and I'll push my finished changes tommorow with more details on my reasoning behind the changes :)

crocodilestick:
Sorry this took so long, through testing I ended up discovering a significant bug to do with the new split library functionality and ended up making a lot of reliability improvements to the ingest system along the way. Thank you for all your work! :)

PR #452, opened 2025-08-01
feat: Add health check endpoint

gelbphoenix:
## Description

This adds a healthcheck at `/health` which allows admins who are using health monitoring tools like Uptime Kuma to have a specific endpoint which they can check.

The response which `/health` sends is:

If everything is ok and the database is accessible (HTTP code: 200 ok):
```json
{
  "status": "ok",
  "uptime": 41895.45,
  "version": "CWA/(VERSION)"
}
```

If the database is not accessible (HTTP code: 503 Service unavailable)
```json
{
  "status": "degraded",
  "uptime": 41895.45,
  "version": "CWA/(VERSION)"
}
```

## Additional Notes

This PR is a modified version of https://github.com/gelbphoenix/autocaliweb/blob/e3534730528b5b420ea02179f71b38b3522d53a7/cps/web.py#L849-L869
Fixes #171

PR #520, opened 2025-08-14
Fix health check database path concatenation

InsideTheVoid:
Updated the health check function to use os.path.join for constructing the database path. 
This resolves the health check always failing due to incorrect path handling.


gelbphoenix:
Definitely recommended. There is a difference between the code of {cwa/acw}_get_library_location() in both Autocaliweb and Calibre-Web Automated which I haven't seen when writing #452.

PR #526, opened 2025-08-15
Remove `.` in ingest_ignored_formats for partial download

Aymendje:
The logic does a `Path(filename).suffix.replace('.', '')` which then compares to the ingest_ignored_formats with a dot, meaning it never matches.


crocodilestick:
Thank you 🙏

Issue #509, opened 2025-08-12
Hardcover not updating


Jondeli:
Hardcover does not seem to udpate when starting a new book or finishing a book with Kobo. Finishing a book does change the read status of the book in CWA itself but it is not updated to Hardcover.

I have updated the book metadata with Hardcover information like the ID and editions, Hardcover metadata fetching works. I have added the API token in the docker env so that part seems OK.

I have also forced an update to Kobo from the user settings. I have also set the API token in the user setting. But that part does not seem to work, it does not reflect any changes to Hardcover side.

There still seems to lack proper documentation for this feature so might be that I have done something wrong so don't consider this a bug atm. The books existed in the shelf before latest update where this feature came but I have updated the metadata to include hardcover things as mentioned so not sure if that can affect.

smathev:
I think the issue is this error, I'm seeing in the logs:

> WARN {py.warnings:109} /app/calibre-web-automated/cps/services/hardcover.py:110: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if len(ids) is not 0:<

Which is happening here in the file:
```
    # TODO Add option for autocreate if missing books instead of forcing it.
    def update_reading_progress(self, identifiers, progress_percent):
        ids = self.parse_identifiers(identifiers)
        if len(ids) is not 0:
```

I've managed to create a fork, and create a pull request for this minor fix. I'm hoping it will be accepted.
https://github.com/crocodilestick/Calibre-Web-Automated/pull/545

smathev:
My merge was pulled into the main branch, so it should be working in the next release.

crocodilestick:
Fixed in v4.0.0

PR #545, opened 2025-08-20
Should fix issue: 509

smathev:
https://github.com/crocodilestick/Calibre-Web-Automated/issues/509

By changing 'is not' to '!=' so syncing reading progress is actually executed instead of the error we're seeing in the logs:
> WARN {py.warnings:109} /app/calibre-web-automated/cps/services/hardcover.py:110: SyntaxWarning: "is not" with a literal. Did you mean "!="?
  if len(ids) is not 0:<


crocodilestick:
Good job <3

smathev:
> Good job <3

Cheers! It's my first clone, fork, merge - and it was even a success :)

PR #569, opened 2025-08-25
First working version of magic shelf functionality. Not yet finished but getting there

crocodilestick:


crocodilestick (commented):
First working version of magic shelf functionality. Not yet finished but getting there

stripeymonkey:
Nice.Looks like a great feature

Olen:
Love this.
I nice addition would be to define one of the Magic Shelfs as the "Default" or "Frontpage" - and have it take the place of the "Books" as the default page.

The reason is that I have a lot of magazines and PDFs of work-related documents. All of these are nicely tagged with e.g "Work" etc. and I already have a "Work" Shelf, but they still fill up the "Books" page.  So to be able to create a Magic Shelf which ignores all the "Work" and "Magazine" tags, would make navigating the standard books more convenient.

PR #580, opened 2025-08-27
Update messages.po

Valenth:
Added several empty fields.
Updating of the terms 'courriels' little used in common language in favor of 'mail'.

2 typos that have slipped into my translations:

line 1862: 'Merci de patienter' instead of 'Merci d'attente'

line 2041: 'Vignettes' instead of 'Vgnettes'


crocodilestick:
Merci!

PR #651, opened 2025-09-18
Fix logger

spezzino:
Fix 
ERROR {cps.editbooks:154} Failed to queue upload for ingest: module 'cps.logger' has no attribute 'log' when uploading books through the web ui


crocodilestick:
Already fixed this but thank you for contributing! :)

Issue #654, opened 2025-09-18
[bug] Lsof command call in ingest script hangs


Buco7854:
**Describe the bug**
When ingesting a book, the function `is_file_in_use` is called.
Inside it there is a call to `lsof`
https://github.com/crocodilestick/Calibre-Web-Automated/blob/main/scripts/ingest_processor.py#L336

When file limits are too high lsof is prone to hanging. see : https://github.com/lsof-org/lsof/issues/281
Current docker image use Ubuntu LTS 24.04 wich ships with lsof 4.95, unfortunatly this bug fix has been released in 4.99.

So current docker images are affected by this bug.

I suggest adding lower file limits to the default docker compose.
Additionally a more recent ubuntu image should work but next LTS will normally release in 2026 so we cannot use one.

**To Reproduce**
Run dev container and upload a book

**Expected behavior**
Book should be ingersted properly

**Configuration(please complete the following information):**
 - OS: DOcker in Debian 13
 - Hardware: Proxmox VM
 - Version: dev branch

PR #655, opened 2025-09-18
Fix: Lsof hanging

Buco7854:
Fix for lsof hanging, see : https://github.com/crocodilestick/Calibre-Web-Automated/issues/654


crocodilestick:
I've just merged a fix that installs version 4.99 of lsof in the image which should fix the issue however if it doesn't I'll add this too :)

Buco7854:
Just tested with latest dev image and it seems it indeed has been fixed. Thank you.

PR #665, opened 2025-09-23
Fixes tags issue on edition search

demitrix:
Fixes edition query that is broken for no known reason, runs fine elsewhere. Also screws up whitespace in this file a bunch, but it works.


demitrix:
Going to fix the whitespace so this can merge properly.

PR #683, opened 2025-09-29
Fixes: Reverse proxy authentication theme setting

### LINKED ISSUE #682: [bug] Reverse proxy authentication: User based theme switching fails
**Describe the bug**
The user based theme switching doesn't work, if the user is logged in via reverse proxy authentication. Instead, the guest / global settings are used. 

The same race-condition could lead to other permission / settings issues, but none further issues have been observed to date.

**To Reproduce**
Steps to reproduce the behaviour:
1. Setup reverse proxy authentication
2. Go to any page
3. Change theme
4. Nothing happens
5. Change global theme
6. New theme is applied

**Existing analysis**
By adding some log line to the cw_login#utils.py#_get_user and the admin.py#before_request functions, the following sequence was identified.

1. get_user is called: 
  has_request_context -> true
  g.flask_httpauth_user -> None
  current_app.login_manager._load_user()
  returns Guest
2. before_request is called
  calls get_user -> Returns as already defined Guest, as g.flask_httpauth_user is still None
  uses global theme settings
3. usermanagement#user_login_required (or similiar method) is called:
  sets g.flask_httpauth_user based on the used user -> g.flask_httpauth_user = 'User'
4. any further get_user calls:
  Return 'User', as g.flask_httpauth_user is no longer None

**Expected behaviour**
g.flask_httpauth_user is set, before any get_user calls to avoid supplying the wrong user (Guest). Or at least before the admin#before_request is called, to avoid this particular theme issue. 

**Configuration:**
 - OS: Ubuntu Server -> Docker container
 - Hardware: x86 server
 - Version: V3.1.4

stadler-pascal:
Fixes #682 by adding a new decorator that sets the g.flask_httpauth_user if reverse_proxy_header_login works. 

Tested in local deployment with the Dockerfile.


Olen:
I think there might be a larger issue at the back of this, and the PR does not seem to resolve it completely.
I use reverse proxy auth, and when I look at the `current_user` object from `from ..cw_login import current_user` I am `Guest` even if I am logged in - and also after applying this PR.

stadler-pascal:
I think the issue with the additional metadata provider is that we can't guarantee the admin method wrapper will be called before any other code requesting the user is executed. 

If this leads to further issues, then there would be the need for a more comprehensive solution: 
* One idea would be to incorporate user management into the cw_login utils. 
* Another option would be to ensure that the user management method is called first, but I don't know the underlying framework well enough to determine if it's possible to call the method directly at startup.

If anyone has any preferences, I can adapt the commit: I just wanted to fix the issue without changing any more than necessary, as I am new to the codebase and the project.

Olen:
I definitely don't have any preferences, and I don't know the code well enough to comment much, I just thought these were related.

Olen:
Tried to apply the latest fix, but it seems like `current_user` is still Guest in the metadata-providers at least.

Maybe there is some other class I should access?
I just need to provide the users email address as part of identification towards a metadata provider I am looking into setting up.

```
from ..cw_login import current_user

class ForrigeBok(Metadata):
    def search(....)
        log.info("User:")
        log.info(current_user)
```
Result:
```
[2025-10-06 16:20:47,683]  INFO {cps.metadata_provider.forrigebok:39} User:
[2025-10-06 16:20:47,683]  INFO {cps.metadata_provider.forrigebok:40} <User 'Guest'>
```

Tried to repoen cwa in a Private Window and log in again with reverse proxy.
The reverse proxy login works, and in admin I am the correct uer:

image

crocodilestick:
@Olen this fix isn't yet in dev or any other image, just confirming you added it manually right?

Olen:
> @Olen this fix isn't yet in dev or any other image, just confirming you added it manually right?

Yes. 
I added the 

```
    @app.before_request
    def _load_reverse_proxy_user():
...
```
to `/app/calibre-web-automated/cps/__init__.py` inside the container and restarted the container.

crocodilestick:
@Olen I just noticed you're testing this in the context of the metadata provider system. The issue you're seeing is a thread context problem.

The metadata search runs in a `ThreadPoolExecutor` (for parallel searches), and Flask's `g` object (where `g.flask_httpauth_user` is stored) is **thread-local** - it doesn't transfer to worker threads. So when your metadata provider accesses `current_user` in the worker thread, it doesn't see the reverse proxy user.

**Quick workaround** for your ForrigeBok provider:

```python
from flask import session
from ..ub import session as ub_session, User

class ForrigeBok(Metadata):
    def search(self, query, generic_cover="", locale="en"):
        # Get user from session instead of current_user (session DOES transfer to threads)
        user_id = session.get('_user_id')
        if user_id:
            user = ub_session.query(User).filter(User.id == user_id).first()
            user_email = user.email if user else None
        else:
            user_email = None
        
        # Use user_email for your metadata provider API call
```

Olen:
Right.  So my issue is not related.  

Tested your alternative, and it seems like "_user_id" is not set in the session either.
But since this seems to be unrelated to what _this_ PR is about, I'll dig a bit more and open another issue if I need to.

PR #685, opened 2025-09-30
Update kobo.py

PulsarFTW:
Adding additional values and changing some default values from false to true. 
This should allow other sync options to work in addition to CWA.

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -1264,7 +1264,7 @@ def NATIVE_KOBO_RESOURCES():
         "kobo_redeem_enabled": "True",
         "kobo_shelfie_enabled": "False",
         "kobo_subscriptions_enabled": "True",
-        "kobo_superpoints_enabled": "False",
+        "kobo_superpoints_enabled": "True",
         "kobo_wishlist_enabled": "True",
         "library_book": "https://storeapi.kobo.com/v1/user/library/books/{LibraryItemId}",
         "library_items": "https://storeapi.kobo.com/v1/user/library",


PulsarFTW:
User on discord:
> This has fixed my OverDrive integration with no impact on Kobo syncing with CWA or Instapaper; I don't use the other integrations, so didn't test those

PulsarFTW:
The Superpoints feature was an oversight! Just fixed that.

So in my research it looks like the default for `sign_in_page` has changed to https://auth.kobobooks.com/ActivateOnWeb at some point since the CW defaults were set-up. However, the default for `registration_page` remains the same at https://authorize.kobo.com/signin?returnUrl=http://kobo.com/ 

This new `sign_in_page` is the default used by [Komga](https://github.com/gotson/komga/blob/bdca990e8290ae879d8bc58839a12c599bf54bc2/komga/src/main/kotlin/org/gotson/komga/infrastructure/kobo/KoboProxy.kt#L299) and [Booklore](https://github.com/booklore-app/booklore/blob/93d496157e1ffc5992514bcea99e5847646e34e8/booklore-api/src/main/java/com/adityachandel/booklore/service/kobo/KoboResourcesComponent.java#L156) too. I think at the moment both work just fine, but thought it's best to update to the new default. In theory it likely doesn't matter as kobosync will be setup after the device is setup and linked to kobo.

Edit: [This reddit user](https://www.reddit.com/r/ereader/comments/1en4mvn/sideload_kobo_before_setting_up_account/) posted their default conf file which also shows the new version (post is from 12 months ago)

PR #697, opened 2025-10-16
Fix: Users with allowed/denied tags see correct value for number of books available to them.

DoubleUynn:
I noticed when playing with guest browsing that even if a guest was only allowed certain tags, they still saw the total number of books in library. I changed from counting the library to counting the number of entries using the pagination object.


crocodilestick (commented):
Excellent work, thank you for contributing! :)

PR #705, opened 2025-10-20
feat(build): Enable local dev build

jgoguen:
Updates the build script to add CLI flags for a local build. This will use the existing checked-out repo instead of cloning the repo fresh, for building images for development and testing before code is ready to push up.

Also adds CLI flags for the different things the script asks for, mostly as a convenience for people like me who might just re-run the build with some things always the same. I added flags for all of them, just seemed silly to only cover some of them, and the default behavior is the same as it is today.

Having flags for values let me shuffle some code around to better reflect what's happening. If a user is provided, for example, there's no point asking for it again or getting the default user.

And I added a .editorconfig file, to avoid whitespace-only changes. I took a guess at line width values.


crocodilestick:
Thank you so much for this! Great work! I've also added a page to the CWA Wiki explaining the script and how to use it, please give it a quick look over if you get the chance :) https://github.com/crocodilestick/Calibre-Web-Automated/wiki/Building-Custom-Docker-Images

jgoguen:
Looks good, thanks!

PR #707, opened 2025-10-20
fix(amazon): Fix Amazon metadata search

### LINKED ISSUE #701: [bug] Can't get Metadata from Amazon
**Describe the bug**

I can't get any Metadata from Amazon.
After some research I found 

[https://github.com/janeczku/calibre-web/issues/2719](https://github.com/janeczku/calibre-web/issues/2719)

[https://github.com/crocodilestick/Calibre-Web-Automated/issues/161](https://github.com/crocodilestick/Calibre-Web-Automated/issues/161)

This solve the 503 Error but sill no Metadata

**To Reproduce**

Fetch Metadata from Amazon

**Expected behavior**

Get Metdata from Amazon

**Configuration(please complete the following information):**
 - OS: Debian
 - Hardware: x64 with Docker
 - Version 3.1.4

jgoguen:
Amazon search for metadata has been returning HTTP 503. This appears to fix that for `amazon.com`. Also makes some changes to the attributes searched for to accommodate Amazon changes.

This may fix #701

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+max_line_length = 80
+tab_width = 2
+trim_trailing_whitespace = true
+
+[*.py]
+indent_size = 4
+max_line_length = 120
+
+[*.{yaml,yml}]
+max_line_length = 120
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -296,6 +296,7 @@ Copyright (C) 2024-2025 Calibre-Web Automated contributors
 - ytilis (1 commits)
 - zelazna (1 commits)
 - zhiyue (1 commits)
+
 # Fork Contributors (crocodilestick/calibre-web-automated)
 
 - crocodilestick (735 commits)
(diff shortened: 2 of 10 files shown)


redskull1986:
Hello,

thank you for the fix. I've tested it and it works, somehow.

I didn't get any ratings.

I've hardcoded the rating for testing:

```
                    try:
                        match.rating = 3
                    except (AttributeError, ValueError):
                        match.rating = 0
                    try:
```

but still no rating.

jgoguen:
I checked the rating fetch and it did need updating I pushed a new commit with that and a couple other fixes, but it still isn't in the UI. Checking with `log.debug(match.rating)` shows it's being set correctly, could it be something isn't making it to the UI or isn't being used?

jgoguen:
Looking at `cps/templates/book_edit.html` it looks like this is where the modal dialog is defined, but I don't see anything that would show the rating.

redskull1986:
I found in `cps/static/js/get_meta.js`

` 
   // Removed as users can't fetch ratings from metadata providers
    // $("#rating").data("rating").setValue(Math.round(book.rating)); 
`

I don't know, when it was changed.

If I uncomment it. CWA don't display the rating, but it save it.

jgoguen:
It might be better to do the UI all in a separate PR then. I know I don't save the rating from metadata fetching, I wouldn't want it to be enabled without any way to disable it.

crocodilestick:
Thank you both for all your work on this! Merging both PRs now :)

PR #737, opened 2025-11-01
chore: polish function add_aliases() in scripts/setup-cwa.sh

lazyusername:
Append text to a file via readable code. The new code will produce the very same output as the previous one.


crocodilestick:
Thank you! :)

PR #740, opened 2025-11-01
#739 Fix content type for mobi/prc book formats

### LINKED ISSUE #739: [Fix] Incorrect content type for mobi/prc book formats
When using OPDS there is incorrect "application/octet-stream" content type for mobi and prc book formats. Correct format is "application/x-mobipocket-ebook"

AlexSat:
Fix #739


crocodilestick:
Thank you for this! :)

PR #747, opened 2025-11-04
fix(amazon): Fix Amazon book search. Again.

jgoguen:
Once again, fix Amazon metadata search returning HTTP 503. I really hope this isn't going to become a cat-and-mouse game with Amazon.


crocodilestick:
Thank you! <3

Issue #721, opened 2025-10-25
[Feature Request] Change Subject to "CONVERT" when sending email to Kindle


pk-666:
**Is your feature request related to a problem? Please describe.**
When I send a PDF to Kindle it's not converted automatically by Amazon to it's proprietary format, allowing sync progress and better adjustment of reading options.

**Describe the solution you'd like**
The email subject should be changed to "CONVERT" isntead of "Send to eReader". With "CONVERT" on subject, Amazon converts the PDF (and other formats).

crocodilestick:
You can now set the send-to-kindle email subject in v4.0.1

Issue #362, opened 2025-04-27
[Feature Request] Email Subject Label


franciscorochacompt:
Hi,
I use Kobo Mail (https://github.com/clisboa/KoboMail).
In my case, it was important that CWA allowed me to configure a label to be added to the subject of messages and thus facilitate the processing of emails in Kobo Mail.
Someting like [KoboMail], [Kobo], [eBook], etc ...
This function already exists natively in Calibre.
Best Regasrds
Francisco

gelbphoenix:
Short question: You know that Calibre-Web (and in extension this project) do has a native [Kobo sync feature](https://github.com/janeczku/calibre-web/wiki/Kobo-Integration), or?

franciscorochacompt:
I know but it is too recent, has extensive prerequisites and does not have PDF support at this time.

franciscorochacompt:
I already know that there is no limit to anything that cannot be overcome with more or less effort ...
The subject of this request/suggestion is to suggest adding to this tool a functionality that exists in Calibre (label in the email subject) given that it is understood that this tool (CWA) is moving towards being a web alternative to Calibre (because in addition to adding functions it also replicates many of the original functions in Calibre).

crocodilestick:
Fixed in v4.0.0

PR #748, opened 2025-11-05
Add per-user customisable email subject for "send to eReader" emails

marauder37:
Use cases:
* setting subject to CONVERT to use Amazon format conversion (#721)
* adding labels to trigger KoboMail processing (#362)

Default subject (localised "Send to eReader") is used if `kindle_mail_subject` is blank.

* add `kindle_mail_subject` attribute to User model and related database schema
* add `subject` parameter to `send_mail()`
* update profile and user admin UI/endpoints
* whitespace linting


crocodilestick:
Thank you!

Issue #726, opened 2025-10-27
[bug] Series information not written into epub when updating metadata


tomried:
**Describe the bug**
For some ebooks, the Series metadata is not written into the library's epub file when editing it in the webinterface. It is however there when downloading the epub via the webinterface.

**To Reproduce**
1. Create empty ebook with epub in calibre with just a title and an author (or see epub in [Testtitle - Testauthor.zip](https://github.com/user-attachments/files/23174167/Testtitle.-.Testauthor.zip))
2. Put the file into the ingest folder
3. Edit the file metadata and add some series (e.g. `Testseries`)
4. In the calibre library folder, the `metadata.opf` file is updated with the series info, however the `Testtitle - Testauthor.epub` is not.
5. When downloading the epub from the webinterface, the series info is in the epub file.

**Expected behavior**
The epub file in the library should also be updated. (It is updated for some files in my library, but not for others. Currently it seems that files with titles with non-Ascii characters are affected, but somehow it does also not work for this empty file.)

**Configuration(please complete the following information):**
 - OS: docker on Debian
 - Hardware: x86 server
 - Version V3.1.4

tomried:
Manually running `ebook-polish -o metadata.opf -U Testtitle\ -\ Testauthor.epub modified.epub` works. It seems that this is not executed, are there any log files which could show why? I cannot see anything in the default log.

tomried:
The issue with the test file seems to be unrelated, it does not work because there is no cover.jpg in the folder:
```
root@1356ba601c54:/app/calibre-web-automated/scripts# python3 /app/calibre-web-automated/scripts/cover_enforcer.py --dir /calibre-library/Testauthor/Testtitle\ \(126\)/
ebook-polish -c "/calibre-library/Testauthor/Testtitle (126)/cover.jpg" -o "/app/calibre-web-automated/metadata_temp/Testauthor/Testtitle/Testtitle - Testauthor.opf" -U "/calibre-library/Testauthor/Testtitle (126)/Testtitle - Testauthor.epub" "/calibre-library/Testauthor/Testtitle (126)/Testtitle - Testauthor.epub"
Traceback (most recent call last):
  File "runpy.py", line 198, in _run_module_as_main
  File "runpy.py", line 88, in _run_code
  File "site.py", line 47, in <module>
  File "site.py", line 43, in main
  File "calibre/ebooks/oeb/polish/main.py", line 421, in main
[...]

PR #750, opened 2025-11-05
fix: apply metadata to file even when no cover.jpg is present

tomried:
Checks if a cover.jpg is present before applying metadata to avoid ebook-polish command failing
Fixes bug in #726


crocodilestick:
Thank you! <3

PR #757, opened 2025-11-09
Fix koreader plugin collision bug

sirwolfgang:
This fixes the issue where it conflicted with the default sync process plugin and moves it to the main menu so it's easier to find. 

At some point, someone can figure out how to make it auto-add OPDS, and the location change will make more sense.


sirwolfgang:
Merging into my bigger PR

PR #764, opened 2025-11-13
Fix KOSync LDAP authentication for users without local passwords

JamesRy96:
## Description
Fixes an issue where LDAP users cannot login to the KOSync plugin unless they first manually set a local password in their CWA profile.

## Problem
The KOSync authenticate_user() function only checked local password hashes.

## Solution
- Added LDAP authentication support to KOSync's authenticate_user() function.
- Verified LDAP authentication works on existing LDAP users who have not manually set passwords in the UI.

### DIFF
--- a/cps/kosync.py
+++ b/cps/kosync.py
@@ -118,11 +118,9 @@ def authenticate_user() -> Optional[ub.User]:
             return user
         if error is not None:
             log.error(f"authenticate_user: LDAP error for user {user.name}: {error}")
-        # If LDAP authentication fails, don't fall back to local password for LDAP-configured systems
-        log.warning(f"authenticate_user: LDAP authentication failed for user: {user.name}")
-        return None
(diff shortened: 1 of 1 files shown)


JamesRy96:
Thanks. 

I'm retrospect, it's probably a lot better experience to fall back to local password if LDAP fails. 

Thinking about how LDAP passwords can require changes after x amount of time. It would be annoying to have to update the login in the plugging whenever that happens.

crocodilestick:
> Thanks.
> 
> I'm retrospect, it's probably a lot better experience to fall back to local password if LDAP fails.
> 
> Thinking about how LDAP passwords can require changes after x amount of time. It would be annoying to have to update the login in the plugging whenever that happens.

Yeah agreed, I changed it so falls back to local password when LDAP fails and merged. Thanks for your contributation! :)

PR #782, opened 2025-11-22
Fix ingest using split db

itsmarcy:
Fixes bug where if a split library is used, this assumes the metadata.db exist alongside it.


crocodilestick:
Excellent spot & work, thank you! :)

Issue #763, opened 2025-11-12
[Feature Request] Allow bypassing certs for SMTP


w1ll1am23:
**Is your feature request related to a problem? Please describe.**
I'm running a local SMTP server that is failing the SSL verification and I can't send emails. 

Socket Error sending e-mail: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate (_ssl.c:1007)

**Describe the solution you'd like**
A way to bypass SSL verification

**Describe alternatives you've considered**
N/A

**Additional context**
I'm running the proton mail bridge via unraid and would like to use it to send emails vs gmail

pihomeserver:
Just got same issue. An option to bypass the certificate check would be great

w1ll1am23:
@pihomeserver I figured out a way to get this working by changing the code in the container.

- Login in to the container
- navigate to `/app/calibre-web-automated/cps/tasks/`
- `nano mail.py`
- edit lines 188 and 198 and swap `create_default_context` with `_create_unverified_context` (not sure which one of these is the important one)
- close and save file
- `rm -rf __pycache__`
- restart the container

I might try to open a PR but I'm not sure how to create a config in the docker compose and get that passed all the way down to this file and I don't have much free time atm so not sure when that is going to happen.

This is going to get reset any time there is an upgrade so you will have to change it again after.

pihomeserver:
Oh nice ! Thank you for searching

I guess that setting an env variable in the docker compose file and test it on both lines should make the job, like
```python
context_method = os.getenv("MAIL_UNVERIFIED_CONTEXT")
context = (not (context_method is None) and context_method == True) and ssl._create_unverified_context()  or ssl.create_default_context()
```
or 
```python
context_method = os.getenv("MAIL_UNVERIFIED_CONTEXT", False)
context = (context_method == True) and ssl._create_unverified_context()  or ssl.create_default_context()
```
I'm not a Python dev so not sure about the syntax and if it's efficient

This way there should not be regression and if we add the variable to true in the yaml then it should work

w1ll1am23:
Thanks @pihomeserver  seems like that is much easier than I thought it would be. Opened a PR.

PR #784, opened 2025-11-23
Add env variable for bypassing certificate verification for SMTP

w1ll1am23:
Adds support for https://github.com/crocodilestick/Calibre-Web-Automated/issues/763

Setting an env variable on the container of `SMTP_ALLOW_UNVERIFIED_SSL` to "true" allows the certificate verification to be disabled for SMTP emails.


crocodilestick:
Thanks for the PR! I made a small tweak to the implementation to improve long-term maintenance. Instead of using ssl._create_unverified_context(), which is a private API (indicated by the underscore) and could theoretically change or disappear in future Python versions, I switched to the standard public API pattern:

```
context = ssl.create_default_context()
context.check_hostname = False
context.verify_mode = ssl.CERT_NONE
```

This achieves the exact same result but uses the official documented way to disable verification. Great work on this! :)

PR #788, opened 2025-11-24
Update Italian translations in messages.po

edxhub:
Updated Italian translations in messages.po file, including corrections and improvements to various strings.


crocodilestick:
Grazie mille!

Issue #715, opened 2025-10-22
[dev] [bug] Testing OIDC with Authentik. Working - but strange issues...


dierochade:
### Environment

I run the dev tag image on docker.  
I migrated from stock cw lately, if this is from interest.

---

### Issue Description

I can login wiht oidc, using authentik, but have two strange issues:

---

#### 1. Login Button Text

Although a custom Login Button text is saved in the dialogue and present after refresh, it is not rendered.  
Reads **"Log in with SSO"**.

---

#### 2. Admin Role Delegation

If I want to use the admin role delegation, there is a very strange behaviour:  
If I set

```
OAuth Scopes
openid profile email groups
```

I get a server 500 error with:

```
raceback (most recent call last):
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 1511, in wsgi_app
response = self.full_dispatch_request()
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 919, in full_dispatch_request
rv = self.handle_user_exception(e)
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 917, in full_dispatch_request
rv = self.dispatch_request()
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 902, in dispatch_request
return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args) # type: ignore[no-any-return]
File "/lsiopy/lib/python3.13/site-packages/flask_dance/consumer/oauth2.py", line 304, in authorized
token = self.session.fetch_token(
self.token_url, ...<2 lines>... **self.token_url_params, )
File "/lsiopy/lib/python3.13/site-packages/requests_oauthlib/oauth2_session.py", line 406, in fetch_token
self._client.parse_request_body_response(r.text, scope=self.scope)
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
File "/lsiopy/lib/python3.13/site-packages/oauthlib/oauth2/rfc6749/clients/base.py", line 426, in parse_request_body_response
self.token = parse_token_response(body, scope=scope)
~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
File "/lsiopy/lib/python3.13/site-packages/oauthlib/oauth2/rfc6749/parameters.py", line 444, in parse_token_response
validate_token_parameters(params)
[...]

PR #803, opened 2025-12-04
Fix too many redirects with generic OAuth.

dmitryikh:
More context: #715

The problem was two fold:
1. BaseOAuth2Session didn't have access to blueprints. Error: 'NoneType' object has no attribute 'token'
2. The token wasn't propagated correctly to BaseOAuth2Session. BaseOAuth2Session was trying to get token from blueprint.token.

I made this fix on my best effort understanding on how flask/ flask_dance / oauth works. Feel free to reject if if you know better way to fix it.


crocodilestick:
Thanks for this but I've already made a more detailed fix that is being built currently (dev-320). Let me know if you still have issues! :)

PR #807, opened 2025-12-05
Prefix /send_selected/ request with correct path

tmacphail:
Send to eReader links didn't work when behind a reverse proxy when it sent an X-Script-Name header.
This fixes it by prefixing with the getPath() function as used elsewhere.


crocodilestick:
Merged, thank you!

PR #814, opened 2025-12-07
[Fix]: Respect `CWA_PORT_OVERRIDE` in svc-calibre-web-automated/run for s6-notifyoncheck

Nementon:
This update ensures that `s6-notifyoncheck` correctly uses the port defined by `CWA_PORT_OVERRIDE` when it is set, inside the svc-calibre-web-automated/run service.

Even if an invalid port is provided, it does not seems to introduce any functional issues. But, this remains likely unexpected and this PR align the `s6-notifyoncheck` with the actual port configuration used by Calibre-Web-Automated.


crocodilestick:
Great catch! Thank you! :)

PR #820, opened 2025-12-09
Parse folders to chown from dirs.json

tmacphail:
I ran into an issue where /cwa-books-ingest was owned by 911:1001 rather than abc:abc when PUID & PGID are set to something else in a docker container.

This modifies the init to parse dirs.json to add to the list of requiredDirs that need to be owned.


crocodilestick:
Thanks for the fix! I made a small hardening tweak in run:

- kept /calibre-library in the default list to avoid regressions if dirs.json is missing or invalid, and switched parsing to Python (already available) so we don’t rely on jq or a brittle sed regex.

Functionally it’s the same goal—ensuring ingest/library/tmp paths get chowned—just a bit more robust.

Issue #797, opened 2025-12-03
[bug] Kobo Sync fails.


nicesocks:
**Describe the bug**
When initiating the sync from my kobo I get "Sync Failed. Please try again."

This used to work fine..

**Configuration(please complete the following information):**
 - OS: unRaid 7.2
-CWA Latest as well as Dev 

**Additional context**
I enabled the logs and got nothing useful other than

[2025-12-02 21:58:24,151] DEBUG {cps.kobo:1121} Kobo: Received unproxied request, changed request port to external server port
[2025-12-02 21:58:24,151] DEBUG {cps.kobo:1131} Kobo: Received unproxied request, changed request url to http://192.168.1.xxx:yyyy

and in the access.log just this:
:ffff:192.168.1.47 - - [2025-12-02 21:23:43] "GET /kobo/cf3462a87e8e8fe008baddbaXXXdc643/v1/initialization HTTP/1.1" 200 10875 0.035390

I'm on the same local network, the kobo can reach the CWA web interface just fine, so the two are seeing each other.

I've tried deleting and re-creating a new auth token, and have tried forcing a full sync.

In CWA I have "Sync only books in selected shelves with Kobo" checked.
and I have "Proxy unknown requests to kobo store" unchecked.

I have books set up with shelves. I've also tried auth tokens for the Admin user in CWA as well as a more generic user.

But all this was working before, just stopped at some point.

ceruleandeep:
The Kobo's OAuth token is expired. Going into your account settings page on the Kobo might cause a refresh, or signing out of your Kobo account and back in. This token comes from Kobo server. It is not related to your CWA `/kobo/f3462a87e` token or `x-kobo-synctoken` or CWA user sign-in.

To the developers: `/v1/initialization` is a bearer-authenticated endpoint. If the access token is expired it will return 401 with body `{"ResponseStatus":{"ErrorCode":"ExpiredToken","Message":"Expired token"}}` instead of the resources map . `kobo.py` L1113 checks whether the response has a resources map, but it doesn't pass the error to the Kobo or write any log message. Now the following calls to UpgradeOAuthTokenCommand and SyncUserProfileCommand will fail and the Kobo displays the network error popup. 

[...]

Issue #770, opened 2025-11-17
[bug] Hardcover request unauthorized with valid token


brokoli18:
**Describe the bug**
I am trying to integrate CWA with hardcover. I have an API key and have tested that it works using hardcovers GraphQL console. I added the key using a docker env var and even checked it in the CWA user settings however when I start a sync on my kobo my progress never gets updated. When I check the CWA logs I see:

`401 Client Error: Unauthorized for url: https://api.hardcover.app/v1/graphql` 

Again this is despite having a valid token in the config

**To Reproduce**
1. Set up CWA kobo integration with valid hardcover key
2. Start sync on docker device
3. See that reading stats are not synced to hardcover

**Expected behavior**
CWA should sync reading stats to hardcover

**Configuration(please complete the following information):**
 - OS: Docker container, Ubuntu OS on docker host
 - Hardware: x86 Server
 - Version: V3.1.4

ceruleandeep:
This can be caused by not having a Hardcover key for the user. Open the user profile and look in Hardcover API token (above  Kobo Sync Token button).

The Hardcover key in Basic configuration > Enable Hardcover Sync is actually used only for metadata provider. For reading state and annotations sync, must set in user profile.

For reading state sync, there is no different message for 'missing key'! Only unauthorized response when empty key is presented to Hardcover.

brokoli18:
@ceruleandeep Now the hardcover error has dissapeared, however my books/reading process still doesnt appear on hardcover. I have added the token to my profile as you advised and added the `hardcover-slug` identifier to certain books, but when I perform a sync from my kobo no information appears on my hardcover profile. I can see no errors in the logs or any references to hardcover at all. Is there something I am missing?

brokoli18:
I explored the problem further in the linked issue, so this can be closed.

Issue #679, opened 2025-09-28
[bug] Kobo Plus tab network error when using calibre web kobo sync


ATanHermit:
**Describe the bug**
After setting up calibre web kobo sync and messing with my proxy settings, it does sync and download books bt the discover tab does not work and shows "Network error" (it actually only happens on Kobo Plus tab, but it defaults there, when it does not for some reason it works).

**Expected behavior**
I can normally navigate the discover menu, and even Kobo Plus.

**Configuration(please complete the following information):**
Docker compose + traefik reverse proxy.

hernil:
I had some similar troubles with that exact setup. Ended up documenting my issues [here](https://devblog.yvn.no/posts/syncing-kobo-and-calibre-web/). Might help you out 😊

TheDarkSpock:
My Calibre Web is locally hosted, but I have the same issue. On the discover tab, ebooks, OverDrive, and audiobooks all seem to work mostly fine. The Kobo Plus tab gives me a network error. 

Also, the wishlist doesn't work at all.

TerrorSource:
> I had some similar troubles with that exact setup. Ended up documenting my issues [here](https://devblog.yvn.no/posts/syncing-kobo-and-calibre-web/). Might help you out 😊

Did your wife also used Kobo Plus?

hernil:
> > I had some similar troubles with that exact setup. Ended up documenting my issues [here](https://devblog.yvn.no/posts/syncing-kobo-and-calibre-web/). Might help you out 😊
> 
> Did your wife also used Kobo Plus?

She does not.

TerrorSource:
I've enabled "DEBUG" logs in CWA and when my friend tries to use the Kobo Plus functionality on the Kobo i see a lot of extra logs in CWA but this one is pretty related to the issue:

`[2025-11-24 11:21:27,025] DEBUG {cps.kobo:1010} Unimplemented Products Request received: http://calibreweb.server.url/kobo/*tokenhere*/v1/deals (request is forwarded to kobo if configured)`

hernil:
It's been some months but I'm pretty sure that the log line you are seeing could very well be expected and everything working right depending on your setup. It's simply stating that the API call is not something that Calibre-web is supposed to intercept and do anything about so it is forwarding the request to Kobo's servers. 

[...]

PR #821, opened 2025-12-09
Fix unopenable Kobo Plus tab, unrenewable auth token, sync problems due to Hardcover issues

ceruleandeep:
Fixes: 

* Issue https://github.com/crocodilestick/Calibre-Web-Automated/issues/679
* Issue https://github.com/crocodilestick/Calibre-Web-Automated/issues/770
* Issue https://github.com/crocodilestick/Calibre-Web-Automated/issues/797
* Unreported UI glitch when adding book to shelf if user Hardcover token not set
* Observed user confusion about the meaning of "forwarded to kobo if configured"

For #679 we are fortunate that there are related commits in upstream. https://github.com/janeczku/calibre-web/pull/3477/commits/654095054f86dfc39a3152810b8d5f354b53fb81 most relevant. https://github.com/janeczku/calibre-web/pull/3477/ also has other changes to Kobo sync which I checked to see if applicable to CWA.

https://github.com/rrgeorge/calibre-web/commit/96149e4e1f6f00bebfa19544cb06a788c0e9f695#diff-07a86a0f46bee91e3fe947bec8a613581aaa97fb5d1430199bc784844701c10cR1158

I think that this change is not necessary in CWA because it is handled by ProxyFix. We should collect test cases from users on HEAD before changing.

https://github.com/rrgeorge/calibre-web/commit/96149e4e1f6f00bebfa19544cb06a788c0e9f695#diff-07a86a0f46bee91e3fe947bec8a613581aaa97fb5d1430199bc784844701c10cR970

https://github.com/janeczku/calibre-web/pull/3477/commits/654095054f86dfc39a3152810b8d5f354b53fb81

2 of these routes are already in CWA with thanks to @bcrdncola on b33d43404b4a9a9804c621f42a8661839ffa30aa.

I was conservative when adding routes. The only new endpoints needed are `GET /v1/products/featuredforkoboplus/` and `GET /v1/categories/<uuid>/products`. In my humble opinion our understanding and documentation-in-code for Kobo interop is increased by only adding endpoints and methods that are actually used by Kobo. I have never seen POST requests to `/v1/products` or `/v1/categories` paths. I have found that the presence of unused routes and methods creates difficulties in understanding Kobo API.

https://github.com/rrgeorge/calibre-web/commit/96149e4e1f6f00bebfa19544cb06a788c0e9f695#diff-07a86a0f46bee91e3fe947bec8a613581aaa97fb5d1430199bc784844701c10cR1016

It is a good idea to filter Calibre UUIDs from being proxied on `/v1/products/<uuid>/reviews|nextread|recommendations` requests. The idea of matching Calibre books to Kobo Store records is also very smart. But trying to find the Kobo UUID for a user Calibre book is unreliable, and delays the response by ~5 seconds. Kobo makes 3 of these requests synchronously at the end of sync cycle, adding ~15 seconds delay. If there is value in returning Kobo data for Calibre books, the matching should be done outside the sync cycle. It would be nice in future to return Calibre-originated metadata about Calibre books.

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -839,17 +839,7 @@ def HandleStateRequest(book_uuid):
             ub.session.rollback()
             abort(400, description="Malformed request data is missing 'ReadingStates' key")
 
-        if config.config_hardcover_sync and bool(hardcover):
-            # Check if book is blacklisted from reading progress syncing
-            book_blacklist = ub.session.query(ub.HardcoverBookBlacklist).filter(
-                ub.HardcoverBookBlacklist.book_id == book.id
-            ).first()
-
-            if book_blacklist and book_blacklist.blacklist_reading_progress:
-                log.debug(f"Skipping reading progress sync for book {book.id} - blacklisted for reading progress")
-            else:
-                hardcoverClient = hardcover.HardcoverClient(current_user.hardcover_token)
-                hardcoverClient.update_reading_progress(book.identifiers, request_bookmark["ProgressPercent"])
+        push_reading_state_to_hardcover(book, request_bookmark)
 
         ub.session.merge(kobo_reading_state)
         ub.session_commit()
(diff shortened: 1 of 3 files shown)


crocodilestick:
Amazing work, great job on this!

PR #827, opened 2025-12-12
Fix generic oauth redirect loop

Caligatio:
Per the [flask-dance docs](https://flask-dance.readthedocs.io/en/latest/signals.html#flask_dance.consumer.oauth_authorized), the `oauth_authorized` signal handler can return a `Response` object to override the default redirect location. If you peek at the [relevant flask-dance source](https://github.com/singingwolfboy/flask-dance/blob/70f79fd9b633cea7b9742219888fac08069edc22/flask_dance/consumer/oauth2.py#L319-L333), the CWA current signal handler attempts a redirect and then returns False. I believe this causes flask-dance to basically ignore the CWA redirect and instead redirect to the defined `next_url` which [CWA sets to /login/generic/authorized](https://github.com/crocodilestick/Calibre-Web-Automated/blob/main/cps/oauth_bb.py#L681). This is what kicks off the redirection loop.

If the `Response` object is instead returned, `flask-dance` [just uses that instead](https://github.com/singingwolfboy/flask-dance/blob/70f79fd9b633cea7b9742219888fac08069edc22/flask_dance/consumer/oauth2.py#L323).


Caligatio:
Note that this fix is likely applicable to Github and Google logins as well but I have no ability to test those.

jakecelentano:
Tested w/ Authentik SSO & worked!!

  Before: Redirect loop on /login/generic/authorized
  After: OAuth login works correctly, user created and logged in

TY!

jacob-card-howe:
> Tested w/ Authentik SSO & worked!!
> 
> Before: Redirect loop on /login/generic/authorized After: OAuth login works correctly, user created and logged in
> 
> TY!

Can confirm this patch works for Keycloak as well! Fantastic sleuthing, thanks for the fix.

web-josh:
This probably also fixes the redirect loop I had with pocket id.

My fix looks like this, but your approach seems cleaner I guess :) Waiting for the merge...

  ```diff
   --- a/cps/oauth_bb.py
   +++ b/cps/oauth_bb.py
   @@ -41,6 +41,7 @@ os.environ['OAUTHLIB_RELAX_TOKEN_SCOPE'] = '1'

    from flask import session, request, make_response, abort
    from flask import Blueprint, flash, redirect, url_for
   +from werkzeug.exceptions import HTTPException
    from flask_babel import gettext as _
    from flask_dance.consumer import oauth_authorized, oauth_error, OAuth2ConsumerBlueprint
    from flask_dance.contrib.github import make_github_blueprint, github
   @@ -840,6 +841,9 @@ def generic_logged_in(blueprint, token):
            except (InvalidGrantError, TokenExpiredError) as e:
                log.error("OAuth token error in generic_logged_in: %s", e)
                flash(_("OAuth authentication failed: Token validation error. Please try again."), category="error")
                return False
   +        except HTTPException:
   +            # Re-raise HTTP exceptions (including abort() redirects) so Flask handles them
   +            raise
            except Exception as e:
                log.error("Unexpected error in generic OAuth login: %s", e)
                flash(_("OAuth authentication failed due to an unexpected error. Please contact administrator."), category="error")
   ```

vworldat:
Can confirm this also works with [Kanidm](https://github.com/kanidm/kanidm/).

Additional note: this OIDC implementation does not seem to support `PKCE`, so the checks have to be disabled in Kanidm's config (`kanidm system oauth2 warning-insecure-client-disable-pkce cwa`).

jeffonthemove (approved):
Works with Keycloak

crocodilestick:
@Caligatio you are the goat for this <3 here's hoping this is the end to this OAuth saga 🤞 Big thanks to everyone else for testing too!

PR #838, opened 2025-12-18
Add libasound2t64 to Dockerfile as a runtime dependency

jeffonthemove:
libasound2t64 appears to be a runtime dependency.  This change should fix the error message below when trying to launch the CWA container.
```
Failed to import PyQt module: PyQt6.QtWebEngineCore with error: libasound.so.2: cannot open shared object file: No such file or directory
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -209,6 +209,7 @@ RUN \
   nano \
   sqlite3 \
   zip \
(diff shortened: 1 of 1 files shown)


crocodilestick:
Thanks for this! Should help out those Raspberry Pi users who were having issues!

Issue #842, opened 2025-12-23
[bug] 'FOREIGN KEY constraint failed' error when deleting book using desktop Calibre


doxxx:
**Describe the bug**
I'm running the dev image of CWA. I wanted to edit the library using Calibre desktop, so I shutdown CWA and opened the library in Calibre desktop. However, when I try to delete a book using Calibre desktop it fails with a "FOREIGN KEY constraint failed" error (traceback below).

**To Reproduce**
1. Shutdown CWA server.
2. Open library in Calibre desktop.
3. Attempt to delete a book.
4. Calibre displays "FOREIGN KEY constraint failed" error.

**Expected behavior**
I expected to be able to delete books from the library using desktop Calibre while the CWA server is not running.

**Configuration:**
 - OS: Synology DSM 7.2.2 update 5
 - Hardware: Synology NAS
 - CWA 3.2.0 dev 331
 - Calibre desktop 8.16.2

**Additional context**
I can still delete books without error through CWA. I tried using the check library function in Calibre but it said everything was fine. I am NOT running CWA and Calibre at the same time.

The Calibre library files (database + books) are hosted on the NAS. I access the library in Calibre desktop via a SMB network share that is mapped to a drive letter in Windows. I have not tried Calibre desktop under Linux yet.

```
Traceback (most recent call last):
  File "calibre\gui2\actions\delete.py", line 54, in delete_one
  File "calibre\db\legacy.py", line 462, in delete_book
  File "calibre\db\cache.py", line 85, in call_func_with_lock
  File "calibre\db\cache.py", line 2224, in remove_books
  File "calibre\db\backend.py", line 2295, in remove_books
  File "calibre\db\backend.py", line 1159, in executemany
  File "C:\t\t\apsw-42s16m2x\src\cursor.c", line 240, in resetcursor
apsw.ConstraintError: ConstraintError: FOREIGN KEY constraint failed
```

ceruleandeep:
Thank you for providing version numbers so that others can try replicating.

About shutdown: after you shut down CWA check for `metadata.db-wal`. If it exists, there is an open connection to the database. If it exists and is not zero-length, there are uncheckpointed transactions. In either case, not properly shut down.

I have observed that shutting down CWA via admin panel does not achieve a shutdown because s6 starts the service again. Certainly, stopping the container is not "shutdown" and likely there are transactions in WAL.

Suggest:

- shutdown via admin panel before stopping container. Stopping container is not a shutdown.
[...]

PR #848, opened 2025-12-29
fix: cascade deletes for progress syncing tables

doxxx:
The tables created for progress syncing have foreign key references to other tables (`books` and `user`). If foreign key constraints are enabled when connecting to the database (as is so in Calibre Desktop), deleting a book or user will fail unless the referencing child rows in the progress syncing tables have also been deleted.

One way to achieve this is to create the progress syncing tables, `book_format_checksums` and `kosync_progress`, with the `ON DELETE CASCADE` clause added to the foreign key reference. This will automatically delete child rows in these tables when their corresponding parent row in in `books` or `user` is deleted. In other words, book checksums and user sync progress will be deleted when the corresponding book or user is deleted.

Fixes: #842

### DIFF
--- a/cps/render_template.py
+++ b/cps/render_template.py
@@ -5,7 +5,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 # See CONTRIBUTORS for full list of authors.
 
-from flask import render_template, g, abort, request, flash
+from flask import render_template, g, abort, request, flash, current_app
 from flask_babel import gettext as _
 from flask_babel import get_locale
 import polib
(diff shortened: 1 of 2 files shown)


doxxx:
I haven't been able to run tests for this yet because I don't have Docker setup (and it's a bit tricky on Bazzite).

jeffonthemove:
It looks like this PR would fix it for new users that don't have these tables in their metadata.db.
Since sqlite doesn't support altering FK constraints, existing users will need to follow the instructions on https://www.sqlite.org/lang_altertable.html#otheralter to create a new table with the cascading FK constraint.  I think we should either:

- Update docs to walk users through exactly what commands to run to perform this migration, or
- Update the code to detect this and automatically perform those steps.  It looks like we can do this within a database transaction, so I think this is the better solution.

doxxx:
> * Update the code to detect this and automatically perform those steps.  It looks like we can do this within a database transaction, so I think this is the better solution.

I'll look into this.

doxxx:
While writing tests for this, I've discovered that if foreign key constraints are not enabled (`PRAGMA foreign_keys = ON`) then the cascaded deletion does not actually occur.

Do you think CWA should enable foreign key constraints when establishing the SQLite connection?

doxxx:
@crocodilestick Any thoughts on this?

amadib:
Will this fix CWA and Calibre on delete/merge?

PR #851, opened 2025-12-30
Updated Polish translation.

FilipMakowski:
I have made an effort to update the Polish translation where it was missing the translation strings, they were incomplete or plain wrong, or where it made sense to rephrase them. I did support myself with AI tools, but I made a lot of effort to review the results and I rephrased things to sound far more naturally. I have NOT checked the locale file against the build, so I just assume that CI/CD will confirm it for me.


crocodilestick:
Bardzo dziękuję!

Issue #479, opened 2025-08-05
Kosync Plugin Not Using LDAP for Authentication


JamesRy96:
**The KOSync plugin will fail to authenticate a LDAP user if a password is not set inside their CWA profile**

**To Reproduce**
1. Import an LDAP user.
2. Install the KoSync plugin on a device.
3. Set CWA server in KoSync.
4. Attempt to login using the users LDAP credentials.

**Expected behavior**
CWA verifies the users credentials against the LDAP source, logins user into plugin.

**Configuration(please complete the following information):**
 - OS: Ubuntu, running in Docker container. 
 - Hardware: x64 server.
 - Version 3.1.2
 
**Workaround**
1. Login to CWA.
2. Open user settings.
3. Set a password in the password field and save.
4. Login to KoSync plugin.

Other features such as the OPDS catalog currently connect to LDAP and verify user credentials.

charredchar:
I'm having this same issue.
I wanted to add that the log shows:
`WARN {cps.kosync:117} authenticate_user: Password mismatch for user: char`
It does work for the local admin account I have saved as a fallback account.

sirwolfgang:
PRs Welcome

JamesRy96:
Forgot I logged this issue a few months ago. 

Submitted PR #764 with a fix.

charredchar:
> Forgot I logged this issue a few months ago.
> 
> Submitted PR [#764](https://github.com/crocodilestick/Calibre-Web-Automated/pull/764) with a fix.

Thanks so much for the work! I wouldn't be able to do anything to help fix the problem myself unfortunately so I'm really happy you were able to resolve it.
I don't mind waiting for the PR to get merged and CWA updated with it. But if that's going to take awhile (as this is the last thing keeping me from making the service live) I also wouldn't mind testing it if that's needed. I assume I just need to update the `kosync.py` file on my instance? If I do that would my instance still be able to update properly on the next release?

JamesRy96:
> > Forgot I logged this issue a few months ago.
> > 
> > Submitted PR [#764](https://github.com/crocodilestick/Calibre-Web-Automated/pull/764) with a fix.
> 
> Thanks so much for the work! I wouldn't be able to do anything to help fix the problem myself unfortunately so I'm really happy you were able to resolve it.
[...]

PR #857, opened 2026-01-01
Allow header auth for kosync endpoint

jbergler:
Sort of related to #479, this endpoint uses a different auth path than the rest of the app, and the logic to check the header auth if configured is missing. Add it.

### DIFF
--- a/cps/progress_syncing/protocols/kosync.py
+++ b/cps/progress_syncing/protocols/kosync.py
@@ -46,7 +46,7 @@ from werkzeug.security import check_password_hash
 from sqlalchemy import func
 from sqlalchemy.exc import SQLAlchemyError
 
-from ... import logger, ub, csrf, config, constants, services
+from ... import logger, ub, csrf, config, constants, services, usermanagement
 from ...render_template import render_title_template
 from ..models import KOSyncProgress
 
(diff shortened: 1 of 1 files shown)


jbergler:
@crocodilestick you might have other opinions on this, but I wonder if the right solution here is to re-use more of the logic from `usermanagement`.

I get the feeling that `@user_login_required` isn't the right decorator for these endpoints, but it doesn't seem too complicated to replace the auth in this file with the shared logic in some way. Maybe it's a version of `requires_basic_auth_if_no_ano` without the guest logic?

Certainly for now, this PR feels like the quick fix, I've tested it on my instance and it solves my problem.

poupounetjoyeux:
Thanks for your fix @jbergler, I was looking for the exact same thing !

Following your last comment I agree with you that having the same logic everywhere using the `requires_basic_auth_if_no_ano` without guest logic could be really nice

Hope the fix will be merged soon

poupounetjoyeux:
Following this, I also propose the #1004 using base decorators
Tested on my side and working without issue

crocodilestick:
Hey @jbergler , sorry this took a while to get merged, great work though!

PR #860, opened 2026-01-02
updated German translations in messages.po

ManuelDrescher:
-Sentence structure errors corrected

-Existing translations were partially adapted to more common words


crocodilestick:
Vielen Dank!

PR #861, opened 2026-01-02
Update German translation for 'Create a Shelf'

ManuelDrescher:
The translation "Bücherregal erzeugen" is cut off in the web UI to "Bücherregal erzeug".

The word "erzeugen" has been replaced with "anlegen" (same meaning in this context), to prevent this.


crocodilestick:
Danke noch mal!

PR #873, opened 2026-01-04
Update Dutch translations for connection error messages

helmerzNL:
Translated a couple of error messages to Dutch.


crocodilestick:
Dank u wel!

PR #875, opened 2026-01-04
fix: logger.get_logger does not exist, so use log variable

halkeye:


crocodilestick:
Thank you for this! Great spot :)

PR #876, opened 2026-01-05
Send WWW-Authenticate header with 401 responses

mehalter:
401 Responses should send a `WWW-Authenticate` header to fix OPDS compatibility with ereader apps. This is a direct backport of a commit provided to calibre-web recently (https://github.com/janeczku/calibre-web/pull/3499)

I have also added a co-author to the commit as to give credit to the original author.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -209,6 +209,7 @@ RUN \
   nano \
   sqlite3 \
   zip \
+  gettext \
   libasound2t64 \
   libxtst6 \
   libxrandr2 \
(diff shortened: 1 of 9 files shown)


mehalter:
Also I'm not sure if there is an explicit process for ingesting calibre-web commits into Calibre-Web-Automated. So feel free to let me know if this is not the correct process, I am just looking to use OPDS with my ereaders. Thanks so much in advance!

crocodilestick:
Thanks for this! This is the correct process as CWA is now officially forked from CW but fixes like this that are useful to both are super appreciated :)

mehalter:
Thanks so much! Got this deployed and my e-reader apps are able to pull books over OPDS now! Thanks again for considering the PR and in such a timely fashion! Keep up the great work on this project!

PR #885, opened 2026-01-09
fix: add `--books-path` argument to `generate_book_checksums.py`

SethMilliken:
Enables `generate_book_checksums` to use `config_calibre_split_dir` setting.

The `generate_book_checksums.py` script does not account for the `config_calibre_split_dir` setting and as a result, when that setting is set, it will attempt to regenerate checksums for the entire library on startup every time, always failing to find any of the books in the default `/calibre-library` location

This PR attempts to remedy this situation by adding a new `--books-path` to the `generate_book_checksums.py` script, defaulting to using the `config_calibre_split_dir` value.

I don't know if this approach is desirable, as it adds an additional database access to a different database to that script (with the `/config/app.db` path hardcoded). I'm also uncertain whether an exception attempting to access that setting ought to abort the script, or if it ought to proceed using the `--library-path` as default.

Given these uncertainties, I have not yet updated the `test_generate_checksums.py` tests. Please let me know if this approach is acceptible and you'd like me to go ahead and update those tests.


crocodilestick:
Thank you for all your hard work on this! I took the liberty of finishing off the PR myself so the fixes are in the next release. Great job! :)

SethMilliken:
Awesome! Thank _you_!

PR #886, opened 2026-01-10
Update German translation for shelf creation message

ManuelDrescher:
Correcting last pull request #861

I did a mistake in this pull request, because "Bücherregal anlegen" is also cut-off.

To preserve the character count and maintain the meaning of the words, here is a translation that approximates the meaning without cutting anything off.

Once again, I apologize for the mistake.

### DIFF
--- a/cps/translations/de/LC_MESSAGES/messages.po
+++ b/cps/translations/de/LC_MESSAGES/messages.po
@@ -1623,7 +1623,7 @@ msgstr ""
 
 #: cps/shelf.py:207 cps/templates/layout.html:270
 msgid "Create a Shelf"
-msgstr "Anlage Bücherregal"
+msgstr "Neues Bücherregal"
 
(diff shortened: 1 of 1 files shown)


vworldat:
@ManuelDrescher Why not either "Regal anlegen" (the original is also "Create a shelf", not "... bookshelf") or "Neues Bücherregal"? Both should fit the limit and either wording seems more natural to me than "Anlage Bücherregal".

ManuelDrescher:
@vworldat 
Yes, "Neues Bücherregal" is for sure the best idear^^. Thanks (pull request corrected)

The term "Bücherregal" should generally be retained. I am not a fan of inconsistency...it is basically called "Bücherregal" everywhere.

But yes, sometimes are the simple idears the best idears. Tanks again.

crocodilestick:
Vielen Dank euch!

PR #887, opened 2026-01-10
fix: load reverse proxy user before ensure_db_session

Olen:
Tested locally, and seems to work.


crocodilestick:
Amazing spot! Thank you so much for helping me put this issue to rest :)

PR #915, opened 2026-01-21
Magic shelf kobo sync fix

burakemirsezen:
Magic shelf syncing to kobo was broken.

Now it functions similar to syncing normal shelves. The only problem is that since there is no magic shelf archive, it can't sync deletions, if the magic shelf is deleted in CWA.


crocodilestick:
Awesome work, thank you for contributing! :)

Issue #950, opened 2026-01-30
[bug]: upgrade to v4.0.1 fails with sqlalchemy.exc.InvalidRequestError: This connection has already initialized a SQLAlchemy


Splinter7914:
**Describe the bug**

Upgraded my docker image from `V3.1.4` to `v4.0.1`.  Startup fails with

docker-compose config

```
name: htpc
services:
  cwa:
    container_name: cwa
    dns:
      - 172.19.0.34
      - 192.168.1.63
    environment:
      HARDCOVER_TOKEN: xxxxx
      PGID: "1000"
      PUID: "1000"
      TZ: America/Chicago
    image: crocodilestick/calibre-web-automated:v4.0.1
    networks:
      default: null
    pull_policy: missing
    restart: unless-stopped
    volumes:
      - type: bind
        source: /var/data/docker/calibre-web-automated
        target: /config
        bind: {}
      - type: bind
        source: /htpc/media/calibre-web-automated/ingest
        target: /cwa-book-ingest
        bind: {}
      - type: bind
        source: /htpc/media/calibre-web-automated/library
        target: /calibre-library
        bind: {}
```

Container logs:

```
To support the app dev(s) visit:
Calibre-Web Automated: https://ko-fi.com/crocodilestick/tip
To support LSIO projects visit:
https://www.linuxserver.io/donate/

───────────────────────────────────────
GID/UID
───────────────────────────────────────

User UID:    1000
User GID:    1000
───────────────────────────────────────
[cwa-init] Starting CWA initialization process...
[cwa-init] Verifying essential directories...
[cwa-init] Directory /app/calibre-web-automated verified
[cwa-init] Directory /config verified
[cwa-init] Directory /calibre-library verified
[cwa-init] Directory /app/calibre verified
[cwa-init] Verifying essential files...
[cwa-init] File /app/CWA_RELEASE verified
[cwa-init] File /CALIBRE_RELEASE verified
[cwa-init] Environment verification complete
[cwa-init] Setting up ImageMagick policy...
[cwa-init] Checking for leftover lock files from previous instance...
[cwa-init] No leftover lock files to remove. Ending service...
[cwa-init] Checking for an existing app.db in /config...
[cwa-init] Existing app.db found!
[cwa-init] Setting binary paths in '/config/app.db' to the correct ones...
[cwa-init] Successfully set binary paths in '/config/app.db'!
[cwa-init] CWA-init complete! Service exiting now...
[...]

PR #953, opened 2026-01-30
Fix migration DDL transaction handling on SQLAlchemy 2.x

bnugent:
## Summary
This fixes a startup crash during DB migrations when running under SQLAlchemy 2.x autobegin.

## What changed
- Use engine.begin() in _run_ddl_with_retry to avoid nested transactions after PRAGMA busy_timeout.
- Add a smoke test to guard against regressions in migration DDL transaction handling.

## Why
PRAGMA triggers autobegin in SQLAlchemy 2.x. The previous conn.begin() then raised InvalidRequestError, which prevented the user.kindle_mail_subject (and related) migrations from applying and left upgraded installs in a crash loop.

## Testing
- uv run --no-project --with pytest --with requests pytest tests/smoke/test_migration_transaction_smoke.py -q

Refs #950

### DIFF
--- a/cps/ub.py
+++ b/cps/ub.py
@@ -71,12 +71,10 @@ def _run_ddl_with_retry(engine, statements, retries=5, base_delay=0.25):
     last_error = None
     for attempt in range(retries):
         try:
-            with engine.connect() as conn:
+            with engine.begin() as conn:
                 conn.execute(text("PRAGMA busy_timeout=5000"))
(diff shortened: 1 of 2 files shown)


mischavandenburg:
I can confirm that this fixed the issue. Thank you.

crocodilestick:
Thanks for this, now just sorting out the mess Calibre 9 has put everyone in

PR #997, opened 2026-02-01
Updated german translation after Major Upgrade to 4.X

ManuelDrescher:
After the major upgrade to version 4.x, many entries were missing translations (likely due to the new features)...or for other reasons.

Missing translations have been added where available in messages.po.

EDIT:
In the next step, all fuzzy entries were checked and corrected.


crocodilestick:
Vielen lieben Dank! :)

PR #1002, opened 2026-02-01
Some Translations

cebo29:


ManuelDrescher:
I did 100% of all missing strings already in #997 today, including 100% of all fuzzy entries.

With the major upgrade to 4.x it is important, to check and "de-fuzzy" all the fuzzy-enries. Otherwise there will be many needed translations missing in this Major upgrade.

cebo29:
Ok, thanks

PR #1004, opened 2026-02-01
Uses the same authentication decorator on all modules

poupounetjoyeux:
Make decarotar name more reliable
Remove some code duplication regarding anonymous check
Makes KOSync using standard decorators

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -122,7 +122,7 @@ calibre_db.ensure_session()  # Before each request
 Use consistent decorator stacking:
 ```python
 @blueprint.route('/path')
-@login_required_if_no_ano  # Respects anonymous browsing setting
+@user_login_or_anonymous  # Respects anonymous browsing setting
 @limiter.limit("10 per minute")  # Rate limiting when enabled
 def route_handler():
     # CWA custom auth check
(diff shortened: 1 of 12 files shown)


crocodilestick:
Hey, just did a review of this and there are a few issues:

1. KOSync endpoints lost CSRF exemption. With global CSRF enabled, PUT /kosync/syncs/progress will be rejected without a CSRF token, breaking KOReader sync. This is visible at kosync.py:368-470
2. Unauthorized KOSync requests now return the HTTPBasicAuth default response instead of the JSON error format promised by KOSync.
    - basic_auth_required returns auth.auth_error_callback directly, bypassing the KOSync JSON error handlers and the documented {"error": 2001, "message": "Unauthorized"} behavior. 
    - See usermanagement.py:143-166 and KOSync routes in kosync.py:368-470

If you can resolve these i'll take another look

poupounetjoyeux:
Hello @crocodilestick,

Thank you to have take the time starting the review!

I repushed (and rebase) regarding your comments with a dedicated error handler for KOSync
Just deployed my custom image and it seems to works as expected 👍 

Don't hesitate if you see other things to update

poupounetjoyeux:
Hello @crocodilestick 

I just rebased and deployed it locally, all is OK 👌

poupounetjoyeux:
Just rebased

PR #1012, opened 2026-02-01
Fix and improve Spanish translations across the UI.

jaimetur:
Improve Spanish UI translations (fix awkward wording, keep filenames/variables intact, and refine key settings/duplicate-detection strings).


crocodilestick:
¡Muchas gracias!

Issue #994, opened 2026-02-01
[bug] Can't read epub or kepub downloaded on Kobo Clara BW from CWA


Caznadien21:
**Describe the bug**
After fixing the issue #975 I could download epub but i still can't read them on my Kobo Clara BW, it said on my Kobo Error : Impossible to open the document

here the log on Calibre-web-automated : 

[2026-01-31 20:37:30,832]  INFO {cps.kobo:160} Kobo library sync request received
[2026-01-31 20:37:31,225]  INFO {cps.kobo:469} Kobo library metadata request received for book a7388caa-d9c7-4813-aa89-b65dc0a8ab55
[2026-01-31 20:37:31,228] ERROR {cps.epub:49} Could not parse epub metadata of book 187 during kobo sync: XML declaration allowed only at the start of the document, line 2, column 6 (<string>, line 2)
[2026-01-31 20:37:31,294] ERROR {cps.helper:1200} Failed to kepubify metadata for book 187: XML declaration allowed only at the start of the document, line 2, column 6 (<string>, line 2)

It's the same error with all my 187 epubs

Distribution | Version | Détails
Calibre-Web Automated | v4.0.4 | 
Calibre | 9.0.0 | -
Kepubify | v4.0.4 | -

DendyA:
I was getting a very similar error as you, also with version 4.0.4. 

It looks like when an epub is uploaded, either through the web UI or auto-imported, the ```container.xml``` file inside the epub gets corrupted and contains two xml declarations right at the beginning of the file:
``` xml
<?xml version="1.0" encoding="utf-8"?>
<?xml version="1.0"?>
```

If you run the ```CWA Send-To-Kindle EPUB Fixer``` in the admin settings, it should correctly remove the ```<?xml version="1.0"?>``` tag and leave just the one with the encoding defined. Then running Kobo sync should auto-convert and generate a valid kepub that the kobo can download.

Unfortunately a pre-existing kepub version of the ebook won't be corrected by the fixer. It will still contain the double declaration if the conversion happened before the epub was fixed. 

I have been trying to dig through the code to figure out why this second declaration gets inserted but haven't found anything yet.

crocodilestick:
@Caznadien21 this explanation from @DendyA is exactly right, this error is caused by poor quality kepub files. If you convert the file to epub in the edit screen, the fixer should run on it, then if you convert it back to kepub the issue should be gone.

[...]

PR #1013, opened 2026-02-01
kindle_epub_fixer.py: Fixed the adding of duplicate xml declarations

DendyA:
This is related to the discussion in #994 .

When importing the epub (either through the Web UI or through the auto-importer), a double xml declaration would be added. This was first noticed in the ```container.xml``` file but upon further testing I noticed it would also happen to any xml-based files if they did not contain an encoding definition in their xml declaration (the epub's xhtml content files etc).

Below is my understanding of why the bug occurs and my description of what I did to fix it.

The duplicate declaration is added by the ```kindle_epub_fixer.py``` in the ```fix_encoding()``` function. This script is run whenever a file is uploaded. The current logic checks for an xml declaration with an encoding definition:
``` xml
<?xml version="1.0" encoding="UTF-8"?>
```
but if it doesn't find one it will append a declaration with that encoding defined. Regardless of if there is a declaration without an encoding already present. This results in the duplicate declaration. 

To fix this, I have added a case to check for a declaration missing an encoding definition and if it exists replace the entire declaration with one that has the encoding.

As mentioned in the linked ticket, if ```CWA Send-To-Kindle EPUB Fixer``` is run manually on the epub file after it is uploaded it does remove this duplicate declaration. This works because the ```fix_container_xml()``` function is called before the ```fix_encoding()``` function. ```fix_container_xml()``` removes the xml declaration without the encoding and keeps the one with it. Then ```fix_encoding()``` finds no issues with the file and the script continues.


crocodilestick:
Hey, great spot! Though while fixing the suped declarations, you also removed the logic that ensures any whitespace proceeding it was preserved which can also cause issues. I tried to submit a commit to this PR with the revised logic but I don't have permission. This is the correct version:

```python
            elif xml_decl_pattern.match(content):
                # XML declaration present without encoding: replace while preserving leading whitespace.
                content = xml_decl_pattern.sub(r'\1' + new_decl, content, count=1)
                self.fixed_problems.append(
                    f"Added {declared_encoding} to XML declaration missing encoding info in {filename}"
                )
```

If you want to still get contributor credit, just update your PR with the correction and I'll merge :)

DendyA:
Ah yes you are right, thank you for that catch! I have included the update in the PR. I also enabled "Allow edits by maintainers"

PR #1021, opened 2026-02-02
Fixed new fuzzy entries in locale DE

ManuelDrescher:
All new fuzzy entries were fixed in german translation


crocodilestick:
Vielen lieben Dank nochmal :)

PR #1025, opened 2026-02-02
Update German Translation

cebo29:
New strings translated and strings corrected from “Dublette” to “Duplikate” for consistency.


crocodilestick:
Hi guys, I'm adding a fix coming in v4.0.6 that fixes the flash messages so they shouldn't cut off anymore, wait for thee release and see if it's better before changing good translations to be shorter because of the cutting off bug

ManuelDrescher:
Hello,

the switch from "Bücherregal" to "Regal" was made some time ago because many German translations are simply longer than the English original.

And due to character limits in the Web-UI, we had to improvise a bit in some places :-)

That's why, for example, "Regal" is currently used everywhere.

My only concern here was consistency.

If the fix addresses this, it would simplify things considerably.

crocodilestick:
Thank you for your translation contribution!

This PR has been processed as part of a batch consolidation of ~48 translation PRs.
Your translations have been merged into main via PR #1467, which combined all
contributors' work per language.

We really appreciate your help making Calibre-Web Automated accessible to more
people around the world. Please feel free to review the updated translations
on the main branch and submit any further improvements!

PR #1074, opened 2026-02-06
feature(cwasync): disable sync actions when there is no active document

SethMilliken:
Disables the push and pull menu items when there is no active document.

Also parenthetically shows why those menu items are disabled (handling the password not being set case as well).

Edit: pulled sorting hint commit into separate PR #1371 ~~Adds a `sorting_hint = "tools"` property to try to ensure the plugin menu location is more stable.~~

### DIFF
--- a/koreader/plugins/cwasync.koplugin/main.lua
+++ b/koreader/plugins/cwasync.koplugin/main.lua
@@ -251,6 +251,9 @@ function CWASync:addToMainMenu(menu_items)
 
                     self.settings.auto_sync = not self.settings.auto_sync
                     self:registerEvents()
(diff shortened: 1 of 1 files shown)


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.117 at [`31bdd5d`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/31bdd5d). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest`. Thanks @SethMilliken.

Issue #1078, opened 2026-02-07
Cannot click or tap SAVE on user edit/create from iPad Safari browser [bug]


caver01:
**Describe the bug**
Cannot create new users on iPad in default Safari browser. Tapping or clicking SAVE does nothing.

**To Reproduce**
1. Go to 'settings: add new user, or settings:edit user'
2. Click on 'SAVE OR CANCEL'
3. nothing happens. Cannot save or cancel edits

**Expected behavior**
Clicking Save should save the user edits and return you to the user list, or settings screen.

**Screenshots**
A screenshot isn’t really needed here. Just try to edit a user and save from iPad OS in Safari.

**Configuration(please complete the following information):**
 - OS: iPadOS
 - Hardware: iPad Pro 13” M4
 - Version Using latest iPadOS here, 26.2.1.

**Additional context**
Funny enough, other buttons to do things like changing metadata etc. and click SAVE—this all works.
Thinking maybe removal of duplicates in popup as the same issue.

SAVE in user edits works in Chrome browser on iPad.

IcyProtocol:
Having the same issue. Similar issue with searching for metadata and trying to delete books. I think this may have something to do with it: 

calibre-web-automated | [2026-02-09 08:07:53, 058] INFO {flask_wtf.csrf:263} The CSRF token is missing.

kdude55:
I get the same issue using safari on macOS Tahoe. I am not having the issue on Firefox on the same Mac. Issue seems to be related to webkit.

ljsimin:
I have the same issue on Safari 18.1.1,  MacOS Sequoia 15.1.1. Works in Chrome.

sethvoltz:
Same, Safari 26.3 on macOS Tahoe 26.3

SethMilliken:
Looks like this is due to Safari not propagating click events to `<div>` elements. Adding `role="button"` is not sufficient, at least not for macOS Safari. I can't see why these shouldn't just be `<button>` elements anyway, so I've [submitted a PR](https://github.com/crocodilestick/Calibre-Web-Automated/pull/1104) with that change.

SethMilliken:
There were a couple of other issues I found underlying this problem that should now be addresed by the same PR. One of the fixes may address some other Safari breakages.

DAdjadj:
March 8 and I also have this issue. You cannot also delete a book.

holzerseb:
[...]

PR #1104, opened 2026-02-12
fix(user-edit): fix form submission for safari #1078

SethMilliken:
Safari does not appear to propagate `click()` for `<div>` elements, so change these to proper `<button>` elements.

Our button behavior is implemented by jQuery functions, so we set `type="button"` on `<button>` elements so that we do not default to the "submit" behavior (c.f. https://www.w3.org/TR/2011/WD-html5-20110525/the-button-element.html#the-button-element). Without this property, that default submit behavior interferes with our `$("#back").click()` jQuery behavior and results in form submission when the cancel button is clicked (even if it does subsequently redirect to the admin page). Presumably we want to remove the possibility of interference for `#user_submit` and `#btndeluser` behavior as well, even though such interference was not observed during testing (and also for consistency).

Edit: Now also addresses an issue in `uploadprogress.js` that was preventing the handlers for our buttons from getting set in `main.js`.

`$("#form-upload").uploadprogress(...)` and `$("#form-upload-format").uploadprogress(...)` rely on the `uploadprogress` function being defined. If that function is not available, the error from attempting to use it prevents subsequent commands in `main.js` from executing,
resulting multiple breakages in Safari.

To fix this, we add a no-op definition of `uploadprogress()` in the Safari exemption clause in `uploadprogress.js` to satisfy the requirement.

### DIFF
--- a/cps/static/js/uploadprogress.js
+++ b/cps/static/js/uploadprogress.js
@@ -17,7 +17,9 @@
     var isSafari = /safari/i.test(ua) && !/chrome|chromium|crios|android/i.test(ua);
 
     if (!$.support.xhrFileUpload || !$.support.xhrFormData || isSafari) {
-        // skip decorating form
+        // Skip decorating form in Safari, but provide a no-op uploadprogress
+        // function so we do not break the jQuery bindings that make use of it.
+        $.fn.uploadprogress = function(){};
         return;
     }
 
--- a/cps/templates/user_edit.html
+++ b/cps/templates/user_edit.html
@@ -550,12 +550,12 @@
     
     <!-- Action Buttons -->
     <div style="margin-bottom: 4rem; display: flex; flex-direction: column; gap: 0.5rem;">
-      <div id="user_submit" class="btn btn-default" style="width: -webkit-fill-available;">{{_('Save')}}</div>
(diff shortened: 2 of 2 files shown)


SethMilliken:
It looks like this is going to require some further refinement. After additional testing, it looks like "Cancel" is not working reliably.

SethMilliken:
With the `uploadprogress.js` and `type="button"` fixes, the Edit User form is now working correctly and reliably for me in macOS Safari, iOS Safari, and macOS Firefox.

sethvoltz:
Interesting that you note Firefox in here as well. Is this a case of Chrome being overly lax with web standards instead of Safari being overly strict?

SethMilliken:
FWIW, the Firefox check was just for potential regressions (and that is my primary browser).

I suspect in this case it's not so much that Safari is being strict _per se_, it's more that treating other elements as buttons is just something that they never implemented because AFAICT, it's a feature above and beyond the spec, so...fair enough.

noahkiss:
+1, would love to see this merged — hitting the same Safari issue (#1206) on a fresh v4.0.6 install where Save silently no-ops across admin/metadata forms. This patch fixes it.

PR #1116, opened 2026-02-16
noop: Add HTML fixture files for upcoming Amazon metadata provider tests

jgoguen:
Adds a set of HTML fixture files for future tests of the Amazon metadata provider. This PR is a no-op.

PR stack:

* #1123
* #1122
* #1121 
* #1120 
* #1119 
* #1116 👈
* `main`

### DIFF
--- a/tests/fixtures/html/amazon_all_fields.html
+++ b/tests/fixtures/html/amazon_all_fields.html
@@ -1,0 +1,62 @@
+<!doctype html>
+<html>
+  <head>
+    <title>Test Amazon Book Detail Page</title>
+  </head>
+  <body>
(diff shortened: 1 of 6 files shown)


jgoguen:
GitHub is dumb and won't allow opening a stack of PRs because I don't have write access to this repo. So as much as this is going to suck, I can either open independent PRs and note dependencies, or I can open one at a time until each one is approved. If I open independent PRs, GitHub will show cumulative changes in each one until earlier PRs are merged.

The end goal here is to minimize the size of each PR and get the Amazon metadata provider refactored and made more general for different regions, as a template for doing other metadata providers.

Issue #882, opened 2026-01-09
[bug] Custom Columns and Identifiers missing from AJAX and OPDS API responses


vijaykalam:
**Description**: Manually added "Custom Columns" (Integer and Text types) are visible in the Calibre-Web UI but are completely absent from the JSON and XML API responses.

**Steps to Reproduce:**

1. Create a custom column in Calibre Desktop (e.g., #word_count as Integer).
2. Restart CWA and verify the data is visible in the book details page of the web UI.
3. Call the API using Postman with proper headers:
       /opds/books/letter/A
       /opds/books/letter/00
       /ajax/book/3

4. Observe the response.

**Expected behavior**
The JSON/XML response should include custom columns fields with data

**Configuration(please complete the following information):**
 - Calibre-Web Automated : v3.1.4
 - Calibre : 8.8.0

crocodilestick:
Fixed in v4.0.0

vijaykalam:
Thank you for looking into this one, I upgraded to latest 4.0.0 version, but it seems there is some issue.

I have added a column namely word_count but 
1) its not returning in OPDS Response
2) in the AJAX response I can see the value but key name is **custom_column_1**

PR #1129, opened 2026-02-18
Expose custom columns as category elements in OPDS feed

vijaykalam:
The issue is part of https://github.com/crocodilestick/Calibre-Web-Automated/issues/882

Added custom column fields to the OPDS feed response using the Atom <category> element. Custom columns from the Calibre database are now exposed in the feed, allowing OPDS clients to read and filter by user-defined metadata.

**Changes**

Included custom column data in the OPDS <category> entries with term, label, and scheme attributes
Mapped Calibre custom column values to standard Atom category format for compatibility with OPDS readers

**Why**
Custom columns can contain useful metadata (e.g., read status, genre tags, source, page count, etc.) that wasn't previously available in the OPDS feed. This enables better filtering and organization in OPDS-compatible e-readers and clients.

### DIFF
--- a/cps/db.py
+++ b/cps/db.py
@@ -437,6 +437,24 @@ class Books(Base):
         return self.timestamp.strftime('%Y-%m-%dT%H:%M:%S+00:00') or ''
 
     @property
+    def custom_extra_fill(self):
+        out = []
+        for col_id in cc_classes:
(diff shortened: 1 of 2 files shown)


vijaykalam:
@crocodilestick : Can you please review this one

Issue #1112, opened 2026-02-15
[bug] Kobo sync times out on large libraries even with shelf-based sync enabled


rfsbraz:
Kobo sync consistently fails on a large library (~13,000 books) because the sync endpoint takes 30-45+ seconds to respond, exceeding the Kobo e-reader's ~30-second HTTP timeout. This happens even when "Only sync books in selected shelves" is enabled with a
   shelf containing only 63 books.

   ## Environment

   - CWA version: v4.0.6
   - Kobo model: Libra Colour (firmware 4.44)
   - Reverse proxy: Traefik
   - Library size: ~13,000 books
   - Kobo sync shelf: 63 books
   - `kobo_only_shelves_sync`: enabled
   - `config_kobo_proxy`: enabled

   ## Evidence

   ### Traefik access logs show every sync request returns 499 (client disconnect) after ~30s:

   ```
   "GET /kobo/.../v1/library/sync?Filter=ALL&..." HTTP/1.1" 499 21 ... 29895ms
   "GET /kobo/.../v1/library/sync?Filter=ALL&..." HTTP/1.1" 499 21 ... 29712ms
   "GET /kobo/.../v1/library/sync?Filter=ALL&..." HTTP/1.1" 499 21 ... 30301ms
   "GET /kobo/.../v1/library/sync?Filter=ALL&..." HTTP/1.1" 499 21 ... 30222ms
   ```

   Requests made from curl (which waits longer) complete successfully with HTTP 200 after 40-80s.

   ### CWA debug logs show each sync step takes 10-13s even for a single changed book:

   ```
   09:39:31.106 - Kobo library sync request received
   09:39:31.205 - Kobo Sync: books last modified (0.1s)
   09:39:44.512 - Kobo Sync: changed entries: 1 (13s to scan!)
   09:39:55.314 - Kobo Sync: selected to sync: 1 (11s)
   09:40:05.344 - Kobo Sync: cover processing (10s)
   09:40:15.821 - Kobo Sync: remaining books to sync: 0 (10s)
   09:40:15.830 - Kobo Sync: changed states: 1
   ```

   **Total: ~45 seconds to sync 1 book.**

   ### The bottleneck is database scanning, not conversion or network

   - All books are pre-converted to kepub (ran `kepubify --update --inplace --calibre`)
   - No on-the-fly conversion happening during sync
   - Each query step appears to scan the full 13K book database rather than being filtered by the sync shelf

   ## Expected behavior

[...]

PR #1144, opened 2026-02-24
Fix Kobo sync failure with newer firmware: add OIDC discovery endpoint and fix OAuth routes

mihneabulu:
## Problem

Newer Kobo firmware (tested on 4.45.23640) performs a full OIDC authentication 
flow before syncing. After `v1/initialization`, the device calls:

    GET /kobo/<token>/oauth/.well-known/openid-configuration

expecting a standard [OpenID Connect discovery document](https://openid.net/specs/openid-connect-discovery-1_0.html).

CWA has no dedicated handler for this path, so it falls through to the catch-all 
`@kobo.route("/oauth/<path:subpath>")` which calls `make_calibre_web_oauth_response()`. 
This returns OAuth tokens instead of a discovery document. The Kobo rejects the 
response and retries indefinitely, resulting in **"Sync failed. Please try again."** 
on the device with no useful error in the logs.

Two additional issues compound the problem:
- `make_calibre_web_oauth_response()` reads `user_id` from `request.get_json()`, 
  which is `None` on GET requests, so it always returns `"user_id": ""` — the Kobo 
  rejects this and loops
- `/oauth/authorize` had no dedicated route, causing 308 redirects that looped

## Changes

**1. Add `HandleOidcDiscovery()`**

Dedicated route for `/oauth/.well-known/openid-configuration` that returns a 
proper OIDC discovery document using `url_for` + `get_auth_token()`, consistent 
with how other endpoints in the file build URLs.

**2. Add `/oauth/authorize` route to `HandleOauthRequest`**

Prevents 308 redirect loops when the Kobo follows the `authorization_endpoint` 
from the discovery document.

**3. Fix `user_id` in `make_calibre_web_oauth_response()`**

Use `current_user.id` instead of reading from the request body, which is always 
empty on GET requests.

## Testing

- CWA v4.0.6
- Kobo firmware 4.45.23640  
- Deployed behind nginx reverse proxy

Before: infinite loop on `oauth/.well-known/openid-configuration`, sync never completes.  
After: full OIDC flow completes, `v1/library/sync` is reached, books sync successfully.

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -1272,7 +1272,7 @@ def make_calibre_web_oauth_response():
         "token_type": "Bearer",
         "expires_in": 3600,
         "scope": content.get("scope", ""),
-        "user_id": content.get("user_id", ""),
+        "user_id": str(current_user.id) if current_user and not current_user.is_anonymous else content.get("user_id", ""),
         # Include legacy field names used by some Kobo requests
         "AccessToken": access_token,
         "RefreshToken": refresh_token,
@@ -1294,8 +1294,28 @@ def HandleAuthRequest():
     return make_calibre_web_auth_response()
 
 
+
+@csrf.exempt
+@kobo.route('/oauth/.well-known/openid-configuration', methods=['GET', 'POST'])
+@requires_kobo_auth
+def HandleOidcDiscovery():
+    base_url = url_for("kobo.HandleOauthRequest",
+                       auth_token=get_auth_token(),
+                       _external=True).rsplit("/oauth", 1)[0]
+    payload = {
+        'issuer': base_url,
+        'authorization_endpoint': base_url + '/oauth/authorize',
+        'token_endpoint': base_url + '/oauth/token',
+        'userinfo_endpoint': base_url + '/oauth/userinfo',
+        'response_types_supported': ['code'],
+        'subject_types_supported': ['public'],
+        'id_token_signing_alg_values_supported': ['RS256'],
+    }
+    return make_response(jsonify(payload))
+
 @csrf.exempt
 @kobo.route("/oauth/token", methods=["GET", "POST"])
+@kobo.route("/oauth/authorize", methods=["GET", "POST"])
 @kobo.route("/oauth/refresh", methods=["GET", "POST"])
 @kobo.route("/oauth/<path:subpath>", methods=["GET", "POST"])
 @requires_kobo_auth


rfsbraz:
This is definitely what I'm experiencing with https://github.com/crocodilestick/Calibre-Web-Automated/issues/1112

jgoguen:
Tested with my Kobo, solves sync issues. 🥳

raclo:
Confirmed: this PR fixes Kobo sync on my Kobo Libra 2.
Before the patch, the device failed to sync and /kobo/<token>/oauth/.well-known/openid-configuration 
returned a token-like JSON (access/refresh token) instead of an OIDC discovery document; /kobo/<token>/oauth also redirected to /login.
After applying the changes from this PR, openid-configuration returns proper OIDC fields (issuer, authorization_endpoint, token_endpoint, etc.) and sync + book downloads work again behind Cloudflare/Nginx Proxy Manager. 
Thanks!

WickedSquirrel:
@crocodilestick will you pull this fix?

crocodilestick:
100%, I'll review it as soon as I have time and thank you for contributing! :)

TJDavis84:
+1 Fixed the kobo sync on 2 of my devices as well

Nymacros:
Do you have an estimated timeline for when this will be resolved, @crocodilestick ?

I’m currently running system version 5.13.241423, and unfortunately the fix doesn’t seem to work in my setup.

Nymacros:
> Do you have an estimated timeline for when this will be resolved, @crocodilestick ?
> 
> I’m currently running system version 5.13.241423, and unfortunately the fix doesn’t seem to work in my setup.

After some additional debugging I was able to get sync working and wanted to share the findings here for others on newer firmware and for @crocodilestick in case it's relevant for the integration.

**Confirmed + Additional fixes required for Kobo firmware 5.13.241423 (Kobo Clara BW)**
_Tested on:_

CWA v4.0.6
Kobo firmware 5.13.241423
Deployed behind Nginx reverse proxy (Nginx Proxy Manager)

The three fixes described in this PR are necessary and correct, confirmed working. However, firmware 5.x introduced two additional issues that required manual fixes:

**4. is_proxied check builds incorrect URLs behind reverse proxy**
kobo.py contains multiple if not current_app.wsgi_app.is_proxied: blocks that build URLs with an explicit port (e.g. https://host:8083/...). Behind a reverse proxy, is_proxied returns False because ReverseProxied.proxied only gets set when HTTP_X_FORWARDED_HOST is present, but CWA uses werkzeug's ProxyFix instead, so this flag never gets set. The Kobo receives URLs with the internal port and rejects them.
_Fix: Replace if not current_app.wsgi_app.is_proxied: with if False: in all affected blocks in kobo.py (affects HandleInitRequest, HandleSyncRequest, HandleMetadataRequest, and get_download_url_for_book). This forces CWA to always use the else branch which builds correct external URLs via url_for(..., _external=True)._

**5. Nginx 502 on v1/library/sync due to large x-kobo-synctoken header**
Firmware 5.x sends a significantly larger x-kobo-synctoken header than 4.45. This exceeds Nginx's default header buffer size, causing a 502 Bad Gateway before the request reaches CWA. The Kobo log shows:
Error transferring .../v1/library/sync?Filter=ALL&DownloadUrlFilter=Generic,Android&PrioritizeRecentReads=true
Error: QNetworkReply::UnknownServerError
502 Bad Gateway
_Fix: Add the following to the Nginx proxy configuration:
large_client_header_buffers 4 32k;
proxy_buffer_size 128k;
proxy_buffers 4 256k;
proxy_busy_buffers_size 256k;_

**Result: After all five fixes, the full sync flow completes successfully on firmware 5.13.241423. Books appear on the device after a forced full sync (via CWA's "Delete sync database" option in the user settings).**

vincent-k:
@Nymacros interresting debug, thanks for your help. However issue 4 looks like an issue on your reverse proxy setup since it affects regular handlers that are also used on older firmwares.

Nymacros:
@vincent-k Thanks for your feedback. I might need to investigate further on this. Anyways, issue 5 should be a different issue from my understanding.

fredericpfisterer:
Hello everybody, when is it going to be released ? My wife asks me everyday 😉

fredericpfisterer:
I built the docker image locally and made my wife happy again, the sync works flawlessly.

lguerard:
Looking forward for this to be merged as I'm having the same error message with no usable logs, so hopefully this will fix it...

new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.25 at [`d830028`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/d830028). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest`. Thanks @mihneabulu.

MUbeira0:
This got a lot of independent confirmation from real users (@jgoguen, @raclo, @TJDavis84) and even got backported into a fork's release (Calibre-Web-NextGen v4.0.25) because people needed it. Closing it seems to have been about running out of steam waiting on review, not a problem with the fix itself.

@mihneabulu would you be open to reopening this, or would it help if someone picked it up from here (rebase + fold in the reverse-proxy header handling that came up in the later comments)? Happy to help push it forward if useful.

mihneabulu:
@MUbeira0 I can open it if you want, but I haven't been keeping up with the changes from the later comments in the next gen repo. It may be better for someone who is more up to date on everything to push it forward.

Issue #1075, opened 2026-02-07
[bug] file permissions changed


petriiczech:
**Describe the bug**
When uploading new books, the rights are not recorded as USER, but as ROOT. The book cannot be edited or deleted.

**Configuration(please complete the following information):**
 - OS: LinuxMint
 - Hardware: x86 desktop
 - Version 22.2
 - CWA 4.0.6

**Additional context**
When tested on another computer everything is fine. Would it help to completely remove the container including the configuration (I have no idea how)?
Or where in the configuration can I change the permissions settings when uploading.

phit42:
#1083

crocodilestick:
Consolidated into #1472 — same root cause (file ownership not set correctly after import operations).

Issue #1128, opened 2026-02-17
[bug] Permissions being forced on ingest folder


bcsteeve:
**Describe the bug**
I saw issue #603 and the problem seemed to be that root was forced as the owner.  I gather that got fixed, but it still forces a umask of 022 (equiv to CHMOD 755).  As the user/owner, I should be able to declare whatever permissions I deem suitable.    In my case, I set a umask of 002 (775), but it is ignored, at least for the ingest folder (I haven't bothered to check the others).

**To Reproduce**
1. Stop the container
2. Manually change ingest folder to CHMOD 775
3. Start the container (ingest folder is forced to 755)

**Expected behavior**
LEAVE MY PERMISSIONS ALONE

bcsteeve:
Update:  The problem is larger than I realized at first.  Having manually fixed the permissions on my ingest folder, I now see that after ingestion, your app creates all the book folders with root:root ownership and 755 permissions.  THIS IS WRONG.  You should respect the uid/gid and umask declarations provided in the compose file and only make educated guesses if the user didn't supply those.  I shouldn't have to manually go in and CHOWN/CHMOD all my folders after your app does its thing.

bcsteeve:
Digging in...

The container is correctly based on ghcr.io/linuxserver/baseimage-ubuntu, and the main app (cps.py) runs properly as user abc.

However, the cwa-ingest-service s6 service runs as root.

In the run script, you have:

```bash
( set -o pipefail
        s6-setuidgid abc inotifywait -m -r ...
```

Only inotifywait is executed as abc.

All other commands in the script — including:

timeout python3 ingest_processor.py

cp

rm

queue handling

status file writes

— are executed as root, since the service itself runs as root.

This results in imported books being created as:

```bash
root:root
755
```

regardless of container PUID/PGID.  That's not ok.

**Recommended Fix**

Drop privileges at the start of the service so the entire script runs as abc.

At the top of the cwa-ingest-service/run file, change:
```bash
#!/usr/bin/with-contenv bash
```
To:
```python
#!/usr/bin/with-contenv bash

# Drop privileges to abc (LinuxServer PUID/PGID user)
if [ "$(id -u)" = "0" ]; then
    exec s6-setuidgid abc "$0" "$@"
fi
```
Then remove the inline s6-setuidgid abc wrapper around inotifywait, since it will no longer be needed.

Result

- Ingest processor runs as abc
- Files respect PUID/PGID
[...]

Issue #1083, opened 2026-02-08
[bug] permissions denied on file epub-fixer.log after importing ebook via folder ingestion


rmcbarreto:
**Describe the bug**
I have done a fresh install of Calibre-Web-Automated using docker compose.

I was importing an ebook, and when I clicked on the "Refresh Library" button, it threw this error:
```
[ingest-processor] Auto-send and metadata functionality available
[ingest-processor] Lock acquired successfully (PID: 1691)
[ingest-processor] Checking if file is ready (timeout: 15 minutes): George R.R. Martin - A Dança Dos Dragões.epub
[ingest-processor] WARN: File did not become ready in time or vanished (after 15 minutes): George R.R. Martin - A Dança Dos Dragões.epub
[ingest-processor] NETWORK_SHARE_MODE=true detected; skipping chown of /calibre-library
[ingest-processor] Skipping delete for ignored/temporary file: George R.R. Martin - A Dança Dos Dragões.epub
[ingest-processor] Lock released (PID: 1691)
[cwa-ingest-service] Successfully processed: /cwa-book-ingest/George R.R. Martin - A Dança Dos Dragões.epub
[library-refresh] Library refresh manually triggered by user...
Traceback (most recent call last):
  File "/app/calibre-web-automated/scripts/ingest_processor.py", line 21, in <module>
    from kindle_epub_fixer import EPUBFixer
  File "/app/calibre-web-automated/scripts/kindle_epub_fixer.py", line 60, in <module>
    file_handler = logging.FileHandler(epub_fixer_log_file, mode='w', encoding='utf-8')
  File "/usr/lib/python3.13/logging/__init__.py", line 1219, in __init__
    StreamHandler.__init__(self, self._open())
                                 ~~~~~~~~~~^^
  File "/usr/lib/python3.13/logging/__init__.py", line 1248, in _open
    return open_func(self.baseFilename, self.mode,
                     encoding=self.encoding, errors=self.errors)
PermissionError: [Errno 13] Permission denied: '/config/epub-fixer.log'
[library-refresh] An unexpected error occurred, check the logs ⛔
```

<details><summary>Docker compose</summary>

```
---
# Docker Compose file source: https://github.com/crocodilestick/Calibre-Web-Automated
services:
  calibre-web-automated:
    image: crocodilestick/calibre-web-automated:latest
    container_name: calibre-web-automated
    restart: unless-stopped
    environment:
      # Only change these if you know what you're doing
[...]

PR #1145, opened 2026-02-25
Modify permission model of cwa-ingest-service

ajn142:
Per suggestion of @bcsteeve in #1128, modified permissions model of entire service to run as abc. Addressed my issue, submitting for review and testing.

To test, clone this PR locally, then add the following line to your compose file:

```
services:
  calibre-web-automated:
    ...
    volumes:
    ...
      - /path/to/local/repo/Calibre-Web-Automated/root/etc/s6-overlay/s6-rc.d/cwa-ingest-service:/etc/s6-overlay/s6-rc.d/cwa-ingest-service
...
```

and restart your container.

### DIFF
--- a/root/etc/s6-overlay/s6-rc.d/cwa-ingest-service/run
+++ b/root/etc/s6-overlay/s6-rc.d/cwa-ingest-service/run
@@ -1,5 +1,10 @@
 #!/usr/bin/with-contenv bash
 
+# Drop privileges to abc (LinuxServer PUID/PGID user)
+if [ "$(id -u)" = "0" ]; then
+    exec s6-setuidgid abc "$0" "$@"
+fi
+
 echo "========== STARTING CWA-INGEST SERVICE =========="
 
 WATCH_FOLDER=$(grep -o '"ingest_folder": "[^"]*' /app/calibre-web-automated/dirs.json | grep -o '[^"]*$')
@@ -286,7 +291,7 @@ if [ -n "$(get_stale_temp_interval_from_db)" ]; then
 fi
 
 ( set -o pipefail
-        s6-setuidgid abc inotifywait -m -r --format="%e %w%f" -e close_write -e moved_to "$WATCH_FOLDER" | \
+        inotifywait -m -r --format="%e %w%f" -e close_write -e moved_to "$WATCH_FOLDER" | \
         while read -r events filepath; do
                 handle_event "$filepath"
         done


ajn142:
May also fix #1075 and #1083, I would appreciate any additional testing folks can offer to validate.

bcsteeve:
Partial success.  It left the user:group alone, but still forced both the ingest folder and the ingested book folder to 755 permissions, despite my 002 (775) UMASK.

bcsteeve:
Ok, I think I spotted the culprit, though I don't have time right now to dig further or test.

I was baffled because I was expecting to find an errant CHMOD somewhere but I didn't.  However, there is liberal use of the "install" command in root/etc/s6-overlay/s6-rc.d/cwa-init/run

And I learned something new...

By default, install -d creates directories with mode 755 unless -m is specified.

What I suggest is that a helper function get installed near the top of the script (after verification section).  Something like:

```
#------------------------------------------------------------------------------------------------------------------------
#  Helper: install directory respecting current umask
#------------------------------------------------------------------------------------------------------------------------

install_dir_umask() {
  local target="$1"
  local owner="$2"
  local group="$3"

  # Get current umask (octal)
  local current_umask
  current_umask=$(umask)

  # Compute directory mode: 0777 - umask
  local dir_mode
  dir_mode=$(printf "%03o" $((0777 & ~8#$current_umask)))

  # Create directory using computed mode
  install -d -m "$dir_mode" -o "$owner" -g "$group" "$target" 2>/dev/null \
    || mkdir -p "$target" 2>/dev/null || true
}
```

Then replace all instances of things like this:

```
install -d -o abc -g abc /some/path
```

with something like this:

```
install_dir_umask /some/path abc abc
```

For example,

```
install -d -o abc -g abc "$INGEST_DIR" 2>/dev/null || mkdir -p "$INGEST_DIR" 2>/dev/null || true
```

becomes simply
```
install_dir_umask "$INGEST_DIR" abc abc
```

 Why This Is the “Correct” Engineering Solution

- Honors container-level umask
- No hard-coded permission assumptions
- Compatible with 002, 022, 027, etc.
- Safe for bind mounts
- Future-proof if UMASK changes
- No behavioral regression for other users

There's going to be more efficient code for this, but hopefully it gets you on the right path.  



Overall, I would say STOP trying to coddle users that don't understand permissions (I assume that's why the dev did it like this) and instead properly guide them in the installation instructions to be sure they include 

      - PUID=[their PUID]
      - PGID=[their PGID]
      - UMASK=002 [or 022 or whatever suits their needs]

in their compose file.

Good luck.

bcsteeve:
On second though to above... my octal math breaks if the environment isn't bash, so don't use that.  The better solution is to not use this sledgehammer install command.  Stick with basic mkdir commands that don't muck around with people's permissions.  I've attached a ready-to-go replacement for root/etc/s6-overlay/s6-rc.d/cwa-init/run (added .txt extension due to limitations of attachments)
[run.txt](https://github.com/user-attachments/files/25539701/run.txt)

phidauex:
For what its worth, the fix suggested in the initial commit is partially working for me - it did allow ingest to start working again.

I still have some abnormal behavior since my calibre library is on an NFS share with root squash, so on my host system it comes in as user media:media (1001:1001). When viewed from inside the cwa container, folders tend to show up as owned by 1001:1001 or abc:1001.

However, as initial functionality is restored, I'm grateful and patient!

PR #1033, opened 2026-02-02
Improve Spanish (es) translations

HaruIjima-kun:
Improves Spanish (es) translations:
- Fix incorrect strings
- Complete missing translations
- Unify tone

PR #1160, opened 2026-03-01
Refine Spanish translations for UI messages

dtorner:
Updated various Spanish translations for consistency and correctness. Needs more work, specially for all those Caps... Thanks for the great software!


HaruIjima-kun:
I've already completed all the missing translations in #1033, corrected all the errors including what you mentioned about periods, and unified the tone of all of them to make it more consistent.

PR #1173, opened 2026-03-05
feat: Shelfmark Feature

gangoke:
Idea for Shelfmark feature (https://github.com/calibrain/shelfmark)
Disclosure: Claude assisted

### **Shelfmark Setting in Basic Config:** 
Enable/disable globally, and set Shelfmark URL/Path. URL/Path can also be set with env SHELFMARK_URL.

image

### **Shelfmark Search:**
 Button which takes the query in search box and opens a new tab to `[ShelfmarkURL]\?q=a+query&content_type=ebook`. (Note: content_type is still in dev branch of Shelfmark) OR if search box is empty, just goes to the ShelfmarkURL without any query. Shelfmark Search is hidden if Shelfmark is disabled globally

image

### **Shelfmark Role:** 
If Shelfmark is enabled globally users with Shelfmark Role will see Shelfmark Search, users without will not. Appears on admin, user_edit, and user_table. Shelfmark Role is hidden/disabled if Shelfmark disabled globally.

image
image
image

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -18,7 +18,6 @@ changelogs
 cwa-wiki
 wiki_images
 .devcontainer
-tmpclaude*
 
 # Test infrastructure (not needed in production image)
 tests/
--- a/.dockerignore
+++ b/.dockerignore
@@ -18,6 +18,7 @@ changelogs
 cwa-wiki
 wiki_images
 .devcontainer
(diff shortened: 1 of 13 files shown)


gangoke:
I completly missed Feature Configuration section in Basic Config. Closing for now.

PR #1174, opened 2026-03-05
fix(ingest): add explicit WAL checkpointing after book imports

n1koo:
## Summary

Add `checkpoint_wal()` to the ingest processor that runs
`PRAGMA wal_checkpoint(PASSIVE)` on `metadata.db`:
- After all writes complete for each imported book
- As a safety-net in the `finally` block at process exit

Logs the checkpoint result tuple (`busy`, `log`, `checkpointed`) for
diagnostics.

## Problem

The WAL file grows unbounded during batch imports. In my testing
(7,400+ book library), the WAL reached 80 MB with only 50 of ~20,000
pages checkpointed. SQLite's auto-checkpoint (every 1,000 pages) uses
PASSIVE mode internally, which cannot advance past active readers.

## What this fixes

Without this change, there are zero explicit `wal_checkpoint` calls
anywhere in the codebase — WAL management is left entirely to SQLite's
auto-checkpoint, which silently fails when persistent readers block it.
This adds checkpoint calls at the correct points in the import pipeline
(after calibredb add, metadata fetch, checksum generation, and
timestamp updates all complete per book), ensuring the WAL is
actively managed rather than left to grow unchecked.

## Known limitation

The Calibre-Web server holds a persistent `StaticPool` connection that
blocks all checkpoint modes equally (PASSIVE, TRUNCATE, and FULL all
checkpoint the same 50 pages in my environment). This PR adds the
necessary checkpoint infrastructure; resolving the `StaticPool` reader
is tracked separately and will make these checkpoints fully effective.

### DIFF
--- a/scripts/ingest_processor.py
+++ b/scripts/ingest_processor.py
@@ -871,6 +871,15 @@ class NewBookProcessor:
             if staged_path.exists():
                 os.remove(staged_path)
 
+    def checkpoint_wal(self):
+        """Explicitly checkpoint the WAL to prevent unbounded growth during batch imports."""
+        try:
+            with sqlite3.connect(self.metadata_db, timeout=5) as con:
+                result = con.execute('PRAGMA wal_checkpoint(PASSIVE)').fetchone()
+                print(f"[ingest-processor] WAL checkpoint: busy={result[0]}, log={result[1]}, checkpointed={result[2]}", flush=True)
+        except Exception as e:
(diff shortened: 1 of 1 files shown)


n1koo:
Closing — same reasoning as #1176. The StaticPool blocking premise was disproven through testing. SQLite already runs PASSIVE auto-checkpoints internally every 1,000 pages, so adding another explicit PASSIVE call won't help when the actual root cause of WAL bloat is still unidentified.

PR #1176, opened 2026-03-05
fix(ingest): use TRUNCATE WAL checkpoint to prevent unbounded WAL growth

n1koo:
## Summary

- Switch WAL checkpoint from `PASSIVE` to `TRUNCATE` mode in the ingest processor
- Add `busy_timeout=5000` so the checkpoint has time to wait for readers to release

## Problem

`PASSIVE` checkpoint cannot advance past active readers. The web server's `StaticPool` connection holds a long-lived read snapshot, which means `PASSIVE` silently skips most pages.

Production evidence from a 7,400-book library:
```
WAL checkpoint: busy=0, log=2486, checkpointed=50
```
2,486 WAL pages accumulated but only 50 were checkpointed. The WAL grows unbounded until the container restarts.

## Fix

`TRUNCATE` mode waits for readers to finish (up to the 5s busy timeout), checkpoints all pages, then truncates the WAL file to zero bytes. If it can't acquire the lock within 5s, it logs a warning and moves on — no errors, no behavior change from the status quo.

## Changes

`scripts/ingest_processor.py` — 1 file, 22 lines added:
- New `checkpoint_wal()` method using `PRAGMA wal_checkpoint(TRUNCATE)` with `busy_timeout=5000`
- Called after each book import completes (success path)
- Called once more at process exit as a safety net

> **Note:** Supersedes #1174 which introduced the checkpoint helper using `PASSIVE` mode.

### DIFF
--- a/scripts/ingest_processor.py
+++ b/scripts/ingest_processor.py
@@ -862,6 +862,9 @@ class NewBookProcessor:
                 except Exception as e:
                     print(f"[ingest-processor] WARN: Failed to adjust timestamps after overwrite import: {e}", flush=True)
 
+            # Checkpoint WAL after all writes for this book are complete
+            self.checkpoint_wal()
+
         except subprocess.CalledProcessError as e:
             print(f"[ingest-processor] {staged_path.stem} was not able to be added to the Calibre Library due to the following error:\nCALIBREDB EXIT/ERROR CODE: {e.returncode}\n{e.stderr}", flush=True)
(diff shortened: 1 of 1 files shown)


n1koo:
Closing — the stated premise (StaticPool holding WAL reader locks) was disproven through testing. TRUNCATE requires a stricter exclusive lock than PASSIVE, so it's unlikely to help when the actual root cause of WAL bloat is still unknown. The real investigation into what holds the reader lock still needs to happen.

PR #1186, opened 2026-03-10
Update Spanish Translation on messages.po

neoblackz:
Uploaded missing translations on Spanish and correct some of the text found that were wrong. All changes made with Loco's free online .po editor.


HaruIjima-kun:
I've already completed all the missing translations in #1033, corrected all the errors including what you mentioned about periods, and unified the tone of all of them to make it more consistent.

PR #1201, opened 2026-03-14
Add currently reading magic shelf

Sheol27:
Since we have the progress from kosync, we can easily track which books we are currently reading

### DIFF
--- a/cps/magic_shelf.py
+++ b/cps/magic_shelf.py
@@ -207,9 +207,25 @@ SYSTEM_SHELF_TEMPLATES = {
     #         ]
     #     }
     # },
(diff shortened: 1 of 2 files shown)


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.77 at [`13b987e`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/13b987e). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest`. Thanks @Sheol27.

PR #1202, opened 2026-03-14
Fix bad Hardcover translations in French

Korri:
Lots of translations where using the literal translation of "hard cover" instead of the website name "Hardcover"


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.163 at [`49c0e8f`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/49c0e8f). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest`. Thanks @Korri.

PR #1221, opened 2026-03-23
Subfolder routing for auto-send users in NewBookProcessor

EuanKerr:
- Add per-user auto-send routing based on ingest subfolder name
- Books placed in /cwa-book-ingest/[username]/ are only sent to the matching CWA user

### DIFF
--- a/scripts/ingest_processor.py
+++ b/scripts/ingest_processor.py
@@ -1011,6 +1011,26 @@ class NewBookProcessor:
                 """)
                 auto_send_users = cur.fetchall()
 
+            # Subfolder routing: if the file was ingested from a subfolder,
+            # only send to the user whose name matches that subfolder.
+            target_username = None
+            try:
+                relative = os.path.relpath(self.filepath, self.ingest_folder)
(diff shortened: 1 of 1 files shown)


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.27 at [`905cda9`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/905cda9b421388ef274f526c22744930397dc33a). Thanks @EuanKerr — drop a book into `/cwa-book-ingest/<username>/` and only that user gets the auto-send.

Heads-up for anyone with an existing subfolder layout: files in any subfolder now go through this filter, so if the first segment doesn't match a CWA username, auto-send is skipped (the book still ingests). If you organize the ingest folder by something other than username, you may want to flatten before pulling.

`docker pull ghcr.io/new-usemame/calibre-web-nextgen:latest` if you want it now. Release notes: https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.0.27.

Issue #1210, opened 2026-03-17
[bug] Koreader Sync doesn't seem to sync progress across different formats of the same ebook


jetpacktuxedo:
**Describe the bug**
As the title says, most of my books have both an `epub` and a `kepub` (I have a kobo device), and nothing seems to allow these to sync to one another.

**To Reproduce**
1. Use two devices to download two different formats of the same book
2. Flip a bit into the book on one device
3. Push progress
4. Switch to the other device using the different format
5. Pull progress
6. Observe that nothing is pulled

**Expected behavior**
CWA clearly understands the progress, as it is displayed in the UI as `KOReader Progress` and that updates correctly when both devices push... but when they go to pull it seems to see the two formats as separate books.

**Screenshots**
See below

**Configuration(please complete the following information):**
 - OS: docker on a debian host
 - Hardware: koreader on android with cwa kosync plugin, xteink x4 running [crosspoint-reader](https://github.com/crosspoint-reader/crosspoint-reader) with their native kosync plugin
 - Version 4.0.6

**Additional context**
`Pull progress from other devices now` on koreader on android always returns `Latest progress is coming from this device`
This always results in only the following two log lines:
```
[2026-03-17 11:50:10,899]  INFO {cps.progress_syncing.protocols.kosync:202} User authenticated successfully: <username>
[2026-03-17 11:50:10,908] DEBUG {cps.progress_syncing.protocols.kosync:289} Found book match: Silver Nitrate (ID 520, format EPUB, checksum vkoreader)
```
`Push progress from this device now` gets me these logs:
```
[2026-03-17 11:52:11,071]  INFO {cps.progress_syncing.protocols.kosync:202} User authenticated successfully: <username>
[2026-03-17 11:52:11,072] DEBUG {cps.progress_syncing.protocols.kosync:680} Updated kosync progress for user 3, document 287e86ec900ed4386a90b989ad2a7b32
[2026-03-17 11:52:11,154]  INFO {cps.progress_syncing.protocols.kosync:699} Saved kosync progress: user=3, document=287e86ec900ed4386a90b989ad2a7b32, prog
ress=0.84%
[2026-03-17 11:52:11,156] DEBUG {cps.progress_syncing.protocols.kosync:289} Found book match: Silver Nitrate (ID 520, format EPUB, checksum vkoreader)
[...]

PR #1225, opened 2026-03-26
fix(kosync): prefer calibre identifier for progress records

SethMilliken:
Documents with different checksums mapped to a common Calibre book currently keep their own distinct KOSyncProgress records (progress records hereafter). This results in inconsistencies in progress updates across different devices.

Example:

Device 1 has document _Some Book_ with checksum 333897. Device 2 has document _Some Book_ with checksum 457883. Device 1 updates progress to 87%. Device 2 retrieves progress, but does not get the update from Device 1 since it was stored under checksum 333897.

This change ensures that we look up the Calibre `book_id` for a given document _before_ looking up the associated progress record, so that we can retrieve the common progress record.

When saving a new progress record, or updating an existing one, we use the Calibre `book_id` (if available) for the record's document identifier.

### DIFF
--- a/cps/progress_syncing/protocols/kosync.py
+++ b/cps/progress_syncing/protocols/kosync.py
@@ -43,7 +43,7 @@ from typing import Dict, Optional, Any, Tuple
 from flask import Blueprint, request, jsonify
 from flask_babel import gettext as _
 from werkzeug.security import check_password_hash
-from sqlalchemy import func
+from sqlalchemy import func, desc
 from sqlalchemy.exc import SQLAlchemyError
 
 from ... import logger, ub, csrf, config, constants, services, usermanagement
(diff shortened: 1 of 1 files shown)


jetpacktuxedo (approved):
Not that my review counts for much, but it looks like this should fix the bug I reported 👍

SethMilliken:
Ah, I hadn't even seen your other issue: https://github.com/crocodilestick/Calibre-Web-Automated/issues/1210 (mentioning here now for the backreference).

new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.29 at [`dad3381`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/dad3381). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest`. Thanks @SethMilliken.

PR #1226, opened 2026-03-26
Do not use `install` to create the ingest directory, use mkdir

antifuchs:
install updates the permissions on the directory, which can be unexpected and unwanted. Instead, this change updates the code to just rely on the fallback `mkdir`, which correctly considers the container's umask and creates the ingest directory with predictable permissions, and only if it should not exist yet.

This ports (a slightly modified to be minimal) change proposed by @bcsteeve in https://github.com/crocodilestick/Calibre-Web-Automated/pull/1145#issuecomment-3957501965; I believe that's all that should be required for the ingest directory to be usable and its permissions to remain untouched, so other external tools can still write to it.

### DIFF
--- a/root/etc/s6-overlay/s6-rc.d/cwa-init/run
+++ b/root/etc/s6-overlay/s6-rc.d/cwa-init/run
@@ -107,7 +107,7 @@ PY
 )
 if [ -n "$INGEST_DIR" ]; then
   # Create directory if missing
-  install -d -o abc -g abc "$INGEST_DIR" 2>/dev/null || mkdir -p "$INGEST_DIR" 2>/dev/null || true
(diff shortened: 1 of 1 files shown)


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.16 at [`7b86b41`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/7b86b41). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest` — same compose, swap the image, restart.

The ingest folder now creates correctly on systems without the `install` utility.

Thanks @antifuchs.

PR #1230, opened 2026-03-26
fix: only delete ingest file after successful import

### LINKED ISSUE #1208: [Feature Request] Don't delete books from ingest until import is successful
**Is your feature request related to a problem? Please describe.**
When adding a file to the ingest, if there are any issues, the book will be deleted even though the import fails.

**Describe the solution you'd like**
Leave the file as is and ideally show a notification in the web ui

**Describe alternatives you've considered**
Deleting the file and showing a notification in the web ui would be helpful, but if I have something downloading a file into the bookdrop directory, I don't love the idea of losing the file entirely.

Another option is to move the file to a failed ingest directory, while annoying to remedy the issue and retry the ingest, it's still much better that losing the file entirely.

burakemirsezen:
Closes #1208 

Changed the return type of add_book_to_library to indicate success. In case it fails, it moves the book into the failed folder.

### DIFF
--- a/scripts/ingest_processor.py
+++ b/scripts/ingest_processor.py
@@ -700,7 +700,7 @@ class NewBookProcessor:
 
 
 
-    def add_book_to_library(self, book_path:str, text: bool=True, format: str="text" ) -> None:
+    def add_book_to_library(self, book_path:str, text: bool=True, format: str="text" ) -> bool:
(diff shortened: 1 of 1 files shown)


burakemirsezen:
Ahhh, I'm dumb. There is already a mechanism that handles this.

PR #1252, opened 2026-04-04
Replace hardcoded paths with configurable variables

BeatLink:
This PR replaces hard coded paths within the source code with variables set in cwa_config.py. These variables can be set using environment variables but default to the existing paths so that the existing setup is not broken. 

By removing these hard coded paths which depend on a docker environment, it will allow calibre-web automated to be installed in non docker environments in the future. I am currently working on a nix flake that depends on this PR.

There is also some whitespace removal and cleanup.


BeatLink:
Bump

PR #1258, opened 2026-04-04
Add per-user OPDS shelf exposure storage

jgoguen:
Add app-db models for per-user OPDS exposure of regular and magic shelves, plus helper functions to read and update those selections.

The change is backend-only and does not alter OPDS responses or UI behavior yet. It lays the groundwork for later filtering and user-facing controls.

*** branch stack ***

- #1258 👈
- #1259
- #1260

*** end branch stack ***

PR #1259, opened 2026-04-04
Enforce per-user OPDS shelf filtering in backend

jgoguen:
Update OPDS authorization and book filtering to use the new per-user exposure
tables for regular and magic shelves. Users have their own selected shelf
exposure settings, which are only applied if they have enabled the per-user
option to expose specific shelves to OPDS.

This PR is limited to backend enforcement and related tests. It should not
impact existing users because there is not yet any UI path to enable the main
OPDS restriction setting.

*** branch stack ***

- #1258
- #1259 👈
- #1260

*** end branch stack ***

### DIFF
--- a/cps/db.py
+++ b/cps/db.py
@@ -8,7 +8,6 @@
 import os
 import re
 import json
-import time
 import threading
 from datetime import datetime, timezone
 from urllib.parse import quote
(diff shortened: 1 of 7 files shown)


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.56 at [`6009fbb`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/6009fbb). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest`. Thanks @jgoguen.

PR #1273, opened 2026-04-07
feat: sync KoReader reading progress to Hardcover

JoelBrenstrum:
Add Hardcover integration for KoReader via a callback/listener pattern:

- kosync.py: Add generic register_progress_listener() hook that fires when progress is updated and matched to a Calibre book
- hardcover_koreader.py: New module with Hardcover-specific logic (blacklist checks, token validation, book lookup, API call)
- main.py: Wire up the integration at app startup

This keeps all Hardcover knowledge outside the KoSync plugin, following the same separation of concerns pattern used elsewhere.

### DIFF
--- a/cps/hardcover_koreader.py
+++ b/cps/hardcover_koreader.py
@@ -1,0 +1,102 @@
+# -*- coding: utf-8 -*-
+# Calibre-Web Automated – fork of Calibre-Web
+# Copyright (C) 2018-2025 Calibre-Web contributors
+# Copyright (C) 2024-2025 Calibre-Web Automated contributors
+# SPDX-License-Identifier: GPL-3.0-or-later
+# See CONTRIBUTORS for full list of authors.
+
+"""
+Hardcover ↔ KoReader integration.
+
+Registers a KoSync progress listener that forwards reading progress to
+Hardcover.  Keeps all Hardcover-specific logic *outside* the KoSync plugin.
+
+Usage (called once at app startup):
+    from .hardcover_koreader import register_hardcover_koreader_sync
+    register_hardcover_koreader_sync()
+"""
+
+from . import calibre_db, config, logger, ub
+from .services import hardcover
+
+log = logger.create()
+
+
+def _on_koreader_progress(user: ub.User, book_id: int, percentage: float) -> None:
+    """Push KoReader reading progress to Hardcover.
+
+    Called by the KoSync progress-listener mechanism whenever a KoReader
+    device pushes progress that was successfully matched to a Calibre book.
+
+    All exceptions are caught and logged so that Hardcover issues never
+    prevent KoReader from syncing progress.
+
+    Args:
+        user: The authenticated CWA user.
+        book_id: The Calibre library book ID.
+        percentage: Reading progress as a percentage (0-100).
+    """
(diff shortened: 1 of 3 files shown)


JoelBrenstrum:
Draft as I haven't has a chance to get this up and running locally.

zb140:
I'm just some rando so take my comments for what they're worth :smile: but I have been trying to get progress sync working both directions between Kobo and KOReader for the last few days, so I've been looking at this code a fair bit.

It seems to me that registration of listeners, broadcasting updates, etc, doesn't really belong to `kosync.py`. I wonder if it would make sense to create some kind of progress sync controller in `cps/progress_syncing` to own that logic. It could eventually be responsible for conflict resolution when updates come in from multiple locations. Then kosync would call that function instead of `update_book_read_status`. In fact, most of `update_book_read_status` could become another listener that turns update broadcasts in to ReadBook entries, the Kobo sync could register a listener to handle its needs, your Hardcover listener could plug into that, etc.  And, other "progress providers" could call the same function to broadcast changes to enable two-way syncing between multiple device types.

JoelBrenstrum:
That seems like a good idea. 

I haven't written much python so I threw claude at the problem and it originally put the hardcover call-out directly into kosync, which I wasn't a fan of. 

Asked it to separate it out and it suggested this eventing idea. 

I don't fully grasp this tracking as originally I thought the kobo sync and the kosync would be quite similar in implementation (and they would both sync to hardcover) but they are quite different.


For my specific use case I've found a direct koreader to hardcover plugin that will keep by books up to date.

jgillman:
Worth noting that this new PR seems to do what @zb140 was suggesting: https://github.com/crocodilestick/Calibre-Web-Automated/pull/1324

zb140:
It looks like that just goes in and puts the hardcover push directly in the kosync code. What I had in mind was something more like this:
https://github.com/zb140/Calibre-Web-Automated/blob/progress/cps/progress_syncing/controller.py#L36

(I've been testing in my for for a couple of weeks now and it's working pretty well for me. It may be about time to clean things up a bit and submit a Draft PR).

frbncis:
> It looks like that just goes in and puts the hardcover push directly in the kosync code. What I had in mind was something more like this:
> https://github.com/zb140/Calibre-Web-Automated/blob/progress/cps/progress_syncing/controller.py#L36
> 
> (I've been testing in my for for a couple of weeks now and it's working pretty well for me. It may be about time to clean things up a bit and submit a Draft PR).

I had played with the idea of refactoring out the hardcover code and building a more generalized device sync pipeline for Kobo and Koreader devices... But in the end, I thought it was a bit to much of an architectural change to offer in as a first PR without input from a maintainer. (Plus, I just wanted to get back to reading 😅)

I'm happy to close abandon/mine in favor of a better solution - I actually didn't see any at the time!

zb140:
> But in the end, I thought it was a bit to much of an architectural change to offer in as a first PR without input from a maintainer. (Plus, I just wanted to get back to reading 😅)

Yeah, this is pretty much exactly the reason my changes are sitting in a fork and not in a PR  😄 I definitely don't intend to barge in here and say "this is the right way to do it". I mainly went the generic route because I knew this PR existed and I wanted to try to be compatible with it.

Issue #1139, opened 2026-02-23
[bug]


apollolabsai:
**Describe the bug**
500 error when updating metadata from hardcopy

Calibre-Web Automated
500 Internal Server Error
The server encountered an internal error and was unable to complete your request. There is an error in the application.
Traceback (most recent call last):
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 1511, in wsgi_app
response = self.full_dispatch_request()
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 919, in full_dispatch_request
rv = self.handle_user_exception(e)
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 917, in full_dispatch_request
rv = self.dispatch_request()
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 902, in dispatch_request
return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args) # type: ignore[no-any-return]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
File "/app/calibre-web-automated/cps/usermanagement.py", line 178, in decorated_view
return login_required(func)(*args, **kwargs)
~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
File "/app/calibre-web-automated/cps/cw_login/utils.py", line 318, in decorated_view
return current_app.ensure_sync(func)(*args, **kwargs)
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
File "/app/calibre-web-automated/cps/web.py", line 2954, in show_book
return render_title_template('detail.html',
entry=entry,
...<6 lines>...
kosync_progress_timestamp=kosync_progress_timestamp,
page="book")
File "/app/calibre-web-automated/cps/render_template.py", line 311, in render_title_template
return render_template(instance=config.config_calibre_web_title, sidebar=sidebar, simple=simple,
accept=config.config_upload_formats.split(','),
magic_shelf_routes=magic_shelf_routes,
duplicate_notification=duplicate_notification,
*args, **kwargs)
File "/lsiopy/lib/python3.13/site-packages/flask/templating.py", line 150, in render_template
return _render(app, template, context)
File "/lsiopy/lib/python3.13/site-packages/flask/templating.py", line 131, in _render
rv = template.render(context)
[...]

Issue #1130, opened 2026-02-19
[bug] 500 error when clicking on a book with undownloaded cover


ki5ck:
**Describe the bug**
Browsing books and clicked on books with a broken image for a cover and got an error. Happened twice for different books. It's intermittent, so bug category may not be correct and it could be my installation.

**To Reproduce**
Could not reproduce; when I went to the previous page and tried again it worked.

**Expected behavior**
Should open book detail page.

**Environment**
Docker using Windows and WSL. Version 4.0.6. New installation (yesterday).

**Error Displayed**
Magic Book Well
500 Internal Server Error
The server encountered an internal error and was unable to complete your request. There is an error in the application.
Traceback (most recent call last):
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 1511, in wsgi_app
response = self.full_dispatch_request()
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 919, in full_dispatch_request
rv = self.handle_user_exception(e)
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 917, in full_dispatch_request
rv = self.dispatch_request()
File "/lsiopy/lib/python3.13/site-packages/flask/app.py", line 902, in dispatch_request
return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args) # type: ignore[no-any-return]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
File "/app/calibre-web-automated/cps/usermanagement.py", line 178, in decorated_view
return login_required(func)(*args, **kwargs)
~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
File "/app/calibre-web-automated/cps/cw_login/utils.py", line 318, in decorated_view
return current_app.ensure_sync(func)(*args, **kwargs)
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
File "/app/calibre-web-automated/cps/web.py", line 2954, in show_book
return render_title_template('detail.html',
entry=entry,
...<6 lines>...
kosync_progress_timestamp=kosync_progress_timestamp,
page="book")
File "/app/calibre-web-automated/cps/render_template.py", line 311, in render_title_template
return render_template(instance=config.config_calibre_web_title, sidebar=sidebar, simple=simple,
[...]

Issue #756, opened 2025-11-07
[bug]


ExHongKonger:
**Describe the bug**

**To Reproduce**
1. Click on a book
2. Go to edit
3. Click on fetch metadata 
4. Select book data
5. Click save
6. See error500 Internal Server ErrorThe server encountered an internal error and was unable to complete your request. There is an error in the application.
Traceback (most recent call last):
File "/lsiopy/lib/python3.10/site-packages/flask/app.py", line 1511, in wsgi_app
response = self.full_dispatch_request()
File "/lsiopy/lib/python3.10/site-packages/flask/app.py", line 919, in full_dispatch_request
rv = self.handle_user_exception(e)
File "/lsiopy/lib/python3.10/site-packages/flask/app.py", line 917, in full_dispatch_request
rv = self.dispatch_request()
File "/lsiopy/lib/python3.10/site-packages/flask/app.py", line 902, in dispatch_request
return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args) # type: ignore[no-any-return]
File "/app/calibre-web-automated/cps/usermanagement.py", line 86, in decorated_view
return login_required(func)(*args, **kwargs)
File "/app/calibre-web-automated/cps/cw_login/utils.py", line 302, in decorated_view
return current_app.ensure_sync(func)(*args, **kwargs)
File "/app/calibre-web-automated/cps/web.py", line 852, in index
return render_books_list("newest", sort_param, 1, page)
File "/app/calibre-web-automated/cps/web.py", line 444, in render_books_list
return render_title_template('index.html', random=random, entries=entries, pagination=pagination,
File "/app/calibre-web-automated/cps/render_template.py", line 207, in render_title_template
return render_template(instance=config.config_calibre_web_title, sidebar=sidebar, simple=simple,
File "/lsiopy/lib/python3.10/site-packages/flask/templating.py", line 150, in render_template
return _render(app, template, context)
File "/lsiopy/lib/python3.10/site-packages/flask/templating.py", line 131, in _render
rv = template.render(context)
File "/lsiopy/lib/python3.10/site-packages/jinja2/environment.py", line 1295, in render
self.environment.handle_exception()
File "/lsiopy/lib/python3.10/site-packages/jinja2/environment.py", line 942, in handle_exception
raise rewrite_traceback_stack(source=source)
File "/app/calibre-web-automated/cps/templates/index.html", line 2, in top-level template code
{% extends "layout.html" %}
[...]

Issue #185, opened 2025-01-14
Install without docker?


richardtheb:
I would love to be able to use CWA, but my host doesn't allow docker installs. Is there a way to install it outside of a Docker container?

CountingBeeps:
As someone who has done this, while technically possible, it is extremely difficult. There are a ton of paths that assume you are using docker. Unless you happen to be on NixOS, I think you might be out of luck.

crocodilestick:
This currently isn't planned as CWA utilizes a lot of bash scripts and S6 processes to function. To convert CWA to pure python would be more trouble than it's worth at this point and bare-metal installations aren't something I'd recommend for this kind of service anyway. All the best!

tecosaur:
@GameDungeon I'm currently trying to get this working on NixOS, did you do any work on this that could be shared?

GameDungeon:
I got this far, https://github.com/GameDungeon/nixpkgs/tree/calibre-web-automated but eventually swapped to docker2nix as it was looking unmaintainable in the long term.

PR #1279, opened 2026-04-10
fix: use subquery loading for Books relationships to prevent DetachedInstanceError

### LINKED ISSUE #1067: [bug] Infrequent errors during metadata updates
**Describe the bug**
When hitting save on metadata, 500 errors frequently arise. The comments field is also frequently missing.

```
sqlalchemy.orm.exc.DetachedInstanceError: Parent instance <Books at 0x753f1a49a510> is not bound to a Session; lazy load operation of attribute 'comments' cannot proceed (Background on this error at: https://sqlalche.me/e/20/bhk3)
```

**To Reproduce**
1. Edit a book
2. Fetch metadata
3. Fill in metadata fields (particularly comments)
4. Hit save.

**Expected behavior**
Book is updated properly without error.

**Screenshots**

**Configuration(please complete the following information):**
 - OS: Windows (Docker)
 - Hardware: x64 Server
 - Version v4.0.6

dbraendle:
## Summary
- Switch all `Books` model relationship attributes from default lazy loading to `lazy='subquery'`
- This prevents `sqlalchemy.orm.exc.DetachedInstanceError` when Jinja2 templates access relationship attributes (series, comments, languages, etc.) after the SQLAlchemy session has been closed

## Problem
Intermittent 500 Internal Server Errors occur when sorting or viewing books. The error is:
```
sqlalchemy.orm.exc.DetachedInstanceError: Parent instance <Books at 0x...> is not bound to a Session; 
lazy load operation of attribute 'series' cannot proceed
```

This affects multiple attributes — `series`, `comments`, `languages` — depending on which template code path is hit.

## Root Cause
The default `lazy='select'` strategy defers loading of relationship data until the attribute is first accessed. If the SQLAlchemy session is closed before the Jinja2 template renders (which accesses these attributes), the lazy load fails with a `DetachedInstanceError`.

## Fix
Setting `lazy='subquery'` on all `Books` relationships ensures the data is loaded eagerly as part of the initial query. This has a minimal performance impact since the data is typically needed anyway for rendering, and completely eliminates the class of `DetachedInstanceError` bugs.

## Related Issues
Fixes #1067, #1130, #1139, #756

## Test plan
- [ ] Sort books by various criteria (date, author, title, series) — no more 500 errors
- [ ] View book detail pages — series, comments, languages all render correctly
- [ ] Verify no performance regression on large libraries (1800+ books tested)

### DIFF
--- a/cps/db.py
+++ b/cps/db.py
@@ -404,15 +404,15 @@ class Books(Base):
     has_cover = Column(Integer, default=0)
     uuid = Column(String)
 
-    authors = relationship(Authors, secondary=books_authors_link, backref='books')
-    tags = relationship(Tags, secondary=books_tags_link, backref='books', order_by="Tags.name")
-    comments = relationship(Comments, backref='books')
-    data = relationship(Data, backref='books')
-    series = relationship(Series, secondary=books_series_link, backref='books')
-    ratings = relationship(Ratings, secondary=books_ratings_link, backref='books')
-    languages = relationship(Languages, secondary=books_languages_link, backref='books')
-    publishers = relationship(Publishers, secondary=books_publishers_link, backref='books')
-    identifiers = relationship(Identifiers, backref='books')
+    authors = relationship(Authors, secondary=books_authors_link, backref='books', lazy='subquery')
+    tags = relationship(Tags, secondary=books_tags_link, backref='books', order_by="Tags.name", lazy='subquery')
+    comments = relationship(Comments, backref='books', lazy='subquery')
+    data = relationship(Data, backref='books', lazy='subquery')
+    series = relationship(Series, secondary=books_series_link, backref='books', lazy='subquery')
+    ratings = relationship(Ratings, secondary=books_ratings_link, backref='books', lazy='subquery')
+    languages = relationship(Languages, secondary=books_languages_link, backref='books', lazy='subquery')
+    publishers = relationship(Publishers, secondary=books_publishers_link, backref='books', lazy='subquery')
+    identifiers = relationship(Identifiers, backref='books', lazy='subquery')
 
     def __init__(self, title, sort, author_sort, timestamp, pubdate, series_index, last_modified, path, has_cover,
                  authors, tags, languages=None):


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.17 at [`cc13678`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/cc13678). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest` — same compose, swap the image, restart.

The DetachedInstanceError class this fix targets reproduces empirically on every relationship attribute (`series`, `comments`, `languages`, `authors`, `data`) on the pre-fix code; with this PR applied, none of them raise. As a bonus, bulk-browse query count drops from ~151 SELECTs to ~10 for a 50-book × 3-relationship-read render — measured in [`tests/integration/test_books_relationship_loading.py`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/tests/integration/test_books_relationship_loading.py) on the canonical empty calibre metadata.db.

Thanks @dbraendle.

new-usemame:
---

Heads-up for anyone applying or reviewing this — we shipped `lazy='subquery'` from this PR on the [`Calibre-Web-NextGen`](https://github.com/new-usemame/Calibre-Web-NextGen) build (a community-maintained CWA build) back in our v4.0.16 and recently found that it can cause intermittent relationship loss under SQLAlchemy 2.0 with gevent + a process-captured session.

**Symptom (reported in [our issue #184](https://github.com/new-usemame/Calibre-Web-NextGen/issues/184) + [#185](https://github.com/new-usemame/Calibre-Web-NextGen/issues/185))**: OPDS feeds intermittently emit books with empty `<summary>`, `<author>`, `<dcterms:language>`, `<category>`. Sequential `curl` against a static library returns different responses. Same problem hits Kobo sync (empty metadata fields) and KOReader cwasync (empty `identifiers` collection so books fail to match by ISBN) — sparser reports there because those clients retry and most users don't inspect the wire payload.

**Root cause** (full trace in [`notes/184-185-DESIGN.md`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/notes/184-185-DESIGN.md) on our side, includes the SQL echo output): `subqueryload` reproduces the parent query's `FROM`/`WHERE`/`ORDER`/`LIMIT` as the inner subselect of its post-load. Under `fill_indexpage`'s two-parent-queries-per-request shape (random panel + main feed) on the process-captured `CalibreDB.session` shared across gevent greenlets, the reproduced subselect can pick up extra outerjoins from prior queries' compile state and miss rows for a subset of parent IDs — those rows then have empty relationship collections.

**Fix on our side** (PR [#188](https://github.com/new-usemame/Calibre-Web-NextGen/pull/188), shipped in v4.0.59): switch the nine `Books` relationships to `lazy='selectin'` instead of `lazy='subquery'`. Same eager-load guarantee (resolves the original `DetachedInstanceError` case from upstream #1067 / #1130 / #1139 just as well), but `selectin` emits one atomic `WHERE parent_id IN (?, ?, ...)` per relationship — no parent-SQL coupling, no compile-state leakage path. SQLAlchemy's docs explicitly call `selectin` out as the modern successor to `subquery` for collection eager loads.

Before fix (v4.0.58 build, 30 sequential `/opds/new` against a 21-book test library): summary/author/category counts varied across iterations. After fix (v4.0.59 build, same library): 100 sequential + 50 parallel requests all returned identical relationship counts.

If this PR moves forward, swapping `subquery` → `selectin` is a one-character-per-line change and might save the next person who applies it from chasing a hard-to-reproduce intermittent. Happy to share any of the SQL traces if useful.

crocodilestick (changes requested):
Hey, thanks for this — the DetachedInstanceError is definitely a real pain point. But I'm a bit worried about the performance impact of setting lazy='subquery' on all 8 relationships globally. Eager-loading all that data on every Book query (even when the view only needs book.title) would blow up query counts pretty badly on a large library — 9 queries per page instead of 1 for data that's often unused.

Would you be open to narrowing this to eager-load only at the specific query sites where detached instances actually occur? joinedload() / subqueryload() at query time is the more targeted fix. Happy to help figure out which views/endpoints are hitting the error if you can share how to repro.

Issue #1149, opened 2026-02-26
[Feature Request] Sort by date added would be nice


raviwarrier:
It would be nice to have a sort function based on added date (both - normal and reverse). This allows the users of my library to know what books were recently added.

Please consider adding this. 

Or, even sorting my book id (the one that calibre gives when adding a book), would be nice. 

Cheers.

PS: I installed this last night and am glad you made the project. Calibre behind KASM was a pain. :)

PR #1280, opened 2026-04-10
fix: set timestamp to current time for newly imported books

dbraendle:
## Summary
- Set the `timestamp` field of newly imported books to the current time after ingest
- Ensures books appear at the top of "Recently Added" views regardless of their publication date

## Problem
When importing books via the ingest folder, `calibredb add` sets the `books.timestamp` field from the EPUB's embedded metadata (typically the publication date). A book published in 2009 but imported today gets a timestamp of 2009-xx-xx, making it invisible in "Recently Added" sorting.

This is confusing for users who expect newly imported books to appear at the top.

## Fix
After `calibredb add` completes and `last_added_book_id` is known, update the book's `timestamp` to `datetime.now()`. This is consistent with the existing overwrite-timestamp logic (lines 859-877) which already updates timestamps for overwritten books.

The existing overwrite logic is preserved unchanged — this patch only adds handling for **new** books.

## Related Issues
Relates to #1149 (Sort by date added)

## Test plan
- [ ] Import a book with an old publication date (e.g., 2009) via ingest folder
- [ ] Verify it appears at the top of "Recently Added" in the web UI
- [ ] Verify overwrite imports still update timestamps correctly
- [ ] Verify books imported via other methods (upload, manual) are unaffected

### DIFF
--- a/scripts/ingest_processor.py
+++ b/scripts/ingest_processor.py
@@ -15,6 +15,7 @@ import shutil
 import sqlite3
 import fcntl
 import threading
+from datetime import datetime
 from pathlib import Path
 
 from cwa_db import CWA_DB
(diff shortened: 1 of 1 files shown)


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.16 at [`f4b3535`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/f4b3535). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest` — same compose, swap the image, restart.

Newly imported books now show today's date instead of the file's mtime / 0001-01-01.

Thanks @dbraendle.

PR #1290, opened 2026-04-18
Run ingest service python workers as abc, not root

haraldpdl:
## Problem

With the documented `PUID=103 PGID=112` configuration, `cps.py` (the web app) correctly runs as the `abc` user via `s6-setuidgid abc` in `svc-calibre-web-automated/run`. Inside `cwa-ingest-service/run`, the `inotifywait` invocation is also prefixed with `s6-setuidgid abc`.

Three other python3 invocations in the same script run as root:

- `python3 watch_fallback.py` on the polling path (line 114)
- `timeout $safety_timeout python3 ingest_processor.py "$queued_file"` in the retry loop (line 158)
- `timeout $safety_timeout python3 ingest_processor.py "$filepath"` on the primary event path (line 216)

Because the actual ingest work (`ingest_processor.py`) runs as root, every auto-imported book lands in `/calibre-library` as `root:root`. The web UI, running as `abc`, then fails to delete these books:

```
Deleting book 66 failed: [Errno 13] Permission denied: '/calibre-library/Author Name/Book Title (66)/metadata.opf'
```

## Fix

Prefix each of those three python3 invocations with `s6-setuidgid abc`, matching the pattern already used for `inotifywait` in the same file.

### DIFF
--- a/root/etc/s6-overlay/s6-rc.d/cwa-ingest-service/run
+++ b/root/etc/s6-overlay/s6-rc.d/cwa-ingest-service/run
@@ -111,7 +111,7 @@ wait_for_stable_file() {
 
 run_fallback() {
         echo "[cwa-ingest-service] Falling back to polling watcher" >&2
-        python3 /app/calibre-web-automated/scripts/watch_fallback.py --path "$WATCH_FOLDER" --interval 5 |
+        s6-setuidgid abc python3 /app/calibre-web-automated/scripts/watch_fallback.py --path "$WATCH_FOLDER" --interval 5 |
         while read -r events filepath; do
                 handle_event "$filepath"
         done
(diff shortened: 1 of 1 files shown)


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.17 at [`3d6c034`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/3d6c034). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest` — same compose, swap the image, restart.

This release ships the patch alongside a one-time chown migration that aligns `/calibre-library` and `/cwa-book-ingest` to `abc:abc` on first boot — without it, users whose libraries had pre-existing root-owned author directories would have hit `Permission denied` the first time the abc-uid ingest worker tried to land a new book in those dirs. The migration is gated by a sentinel marker (idempotent across reboots), and skipped under `NETWORK_SHARE_MODE=true` with logged guidance for the manual run. Eight integration tests in [`tests/integration/test_ingest_setuidgid_permission.py`](https://github.com/new-usemame/Calibre-Web-NextGen/blob/main/tests/integration/test_ingest_setuidgid_permission.py) cover the regression scenario, the idempotent sentinel, the NFS-skip path, and end-to-end ingest after migration.

Thanks @haraldpdl.

PR #1295, opened 2026-04-20
fix(metadata search): Send CSRF token in a way to work with Mobile Safari

jgoguen:
For some reason, iPhone Safari doesn't properly send the X-CSRFToken header. This adds it to the request body for metadata fetch, which is allowing mobile Safari to search for metadata from the Fetch Metadata button when editing a book.

Also proactively adding the header to other requests using fetch().

### DIFF
--- a/cps/static/js/get_meta.js
+++ b/cps/static/js/get_meta.js
@@ -238,7 +238,10 @@ $(function () {
       $.ajax({
         url: getPath() + "/metadata/search",
         type: "POST",
-        data: { query: keyword },
+        data: {
+          query: keyword,
+          csrf_token: $("input[name='csrf_token']").first().val(),
+        },
         dataType: "json",
         success: function success(data) {
           if (data.length) {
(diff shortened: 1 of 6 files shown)


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.54 at [`0026ac5`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/0026ac5). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest`. Thanks @jgoguen.

Yllib:
I confirmed this failure against CWA v4.0.6 in production behind Authelia and a reverse proxy. Clicking **Fetch Metadata** consistently returned `Search error`; the server log showed `flask_wtf.csrf: The CSRF token is missing`, and `/metadata/search` returned HTTP 400 before any metadata provider ran.

Explicitly sending `X-CSRFToken` from the first hidden `csrf_token` input with the metadata AJAX POST resolved the issue immediately. I also applied the same explicit header to the `/metadata/provider/<id>` POST. Metadata results returned normally afterward.

This was not limited to Mobile Safari in my case. The dynamic-token handling in this PR matches the observed failure mode.

PR #1298, opened 2026-04-23
fix(healthcheck): follow redirects so 302 /login is not marked unhealthy

rancur:
## Summary

The Docker `HEALTHCHECK` in the project's root `Dockerfile` uses `curl -f` against `http://localhost:${CWA_PORT_OVERRIDE:-8083}/`. When the user is not authenticated, the app responds to the root URL with a `302` redirect to `/login`. `curl -f` treats any non-2xx status as a failure unless `-L` is passed to follow redirects, so the healthcheck exits non-zero and Docker marks the container `unhealthy` even though it is serving traffic correctly.

This trips orchestrators and auto-healers — Docker Swarm will restart the task, Kubernetes liveness probes derived from the healthcheck will kill the pod, Portainer's auto-redeploy will cycle the container, and Watchtower / healthcheck-to-chat integrations will page operators about a healthy container. Reproducible on any fresh deploy using the upstream `Dockerfile` / `docker-compose.yml` as-is.

## The fix

One-line change to the existing `HEALTHCHECK` directive — add `-L` so `curl` follows the 302 to `/login` (which returns 200) and treat that as healthy. Also added `-s` to silence the progress meter in the check output.

```diff
-  CMD curl -f http://localhost:${CWA_PORT_OVERRIDE:-8083}/ || curl -f -k https://localhost:${CWA_PORT_OVERRIDE:-8083}/ || exit 1
+  CMD curl -fsL http://localhost:${CWA_PORT_OVERRIDE:-8083}/ || curl -fsL -k https://localhost:${CWA_PORT_OVERRIDE:-8083}/ || exit 1
```

No endpoint changes, no new dependencies, no behaviour change for containers that were already healthy — the same root URL is hit first; the only difference is that the 302 response now resolves to a healthy 200 instead of a false failure.

I noticed while digging that a `/health` JSON endpoint already exists (`cps/web.py`, `@web.route("/health")`), so a follow-up could point the healthcheck directly at `/health` for a more semantic check (and to avoid auth-layer work on every probe). Keeping this PR deliberately minimal / backport-friendly — happy to open a follow-up that switches to `/health` if the maintainers prefer.

Context for the earlier #452 ("Add health check endpoint"): that PR was closed-for-rework by its author and the maintainer noted the endpoint landed in dev, so this PR is a strictly smaller companion change that makes today's `curl`-based healthcheck actually pass without waiting for any further refactor.

## Test plan

- [x] `docker build .` against this branch succeeds
- [x] Manually verified `curl -fsL http://localhost:8083/` returns 0 (follows 302 to `/login`, 200 OK) on a running container
- [x] Manually verified `curl -f http://localhost:8083/` (current behaviour) returns non-zero on the same container
- [ ] Docker inspect after start-period shows `"Status": "healthy"` with the patched image
- [ ] Confirm no regression when authenticated (root returns 200 directly — the `-L` is a no-op in that case)

---

Reported as part of Barry-Allen automation work for @itswillcurran.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -298,5 +298,6 @@ VOLUME /calibre-library
 
 # Health check for container orchestration
 # Uses shell form to support environment variable substitution for CWA_PORT_OVERRIDE
+# -L follows redirects so the 302 to /login on the root path is treated as healthy
 HEALTHCHECK --interval=30s --timeout=3s --start-period=120s --retries=3 \
-  CMD curl -f http://localhost:${CWA_PORT_OVERRIDE:-8083}/ || curl -f -k https://localhost:${CWA_PORT_OVERRIDE:-8083}/ || exit 1
+  CMD curl -fsL http://localhost:${CWA_PORT_OVERRIDE:-8083}/ || curl -fsL -k https://localhost:${CWA_PORT_OVERRIDE:-8083}/ || exit 1


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.7 at [`653a516`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/653a516) — `curl -fsL` so the `/ → /login` 302 stops failing healthchecks. Drop-in image for users hitting Swarm/K8s/Watchtower restart loops: `ghcr.io/new-usemame/calibre-web-nextgen:latest`. Mergeable back if upstream resumes. Thanks @rancur.

Issue #639, opened 2025-09-13
404 errors with attempted bare-metal deployment


tecosaur:
Hello,

I'm aware that this is developed with only running as the provided Docker image in mind, but this isn't quite a silver bullet. For instance, on NixOS there are compelling reasons to favour systemd services over containers (better lifetime and state management, for one).

So, I've been having a look at #122 and #185 to try to work out how best to do this. After replacing all of the hardcoded docker root paths (e.g. `/app`) with either static remaps (to the NixOS store) or relative paths, I've managed to get something which seems to *almost* be working: https://code.tecosaur.net/tec/golgi/commit/11a14e374cddda8b5b1f2f9eb14af3061902d10c.

The one issue I seem to be having is 404s with book cover images and the web reader. I can upload books without issue, ingest them, edit metadata, and download them. However, cover art URLs (e.g. `/cover/3/og?c=1757758233`) and the `.epub` fetched by the web reader (e.g. `/show/3/epub/file.epub`) result in 404 errors.

Looking at the process output, I can't see any signs of what the issue might be.

```
[2025-09-13 10:09:35,164] DEBUG {cps.services.worker:80} Add Task for user: admin - Upload The Player of Games - Iain M. Banks.epub
[library-refresh] Library refresh manually triggered by user...
[2025-09-13 10:09:47,184] DEBUG {cps.gdriveutils:72} Cannot import pydrive, httplib2, using gdrive will not work: No module named 'pydrive'
[ingest-processor] GDrive functionality available
[2025-09-13 10:09:47,192] DEBUG {cps.services:15} Cannot import goodreads, showing authors-metadata will not work: No module named 'goodreads'
[ingest-processor] Auto-send and metadata functionality available
[ingest-processor] WARN: 'lsof' command not found. Cannot reliably check if file is in use. Proceeding with caution.
[ingest-processor]: No conversion needed for new_1_20250913_100935_164028_The_Player_of_Games_-_Iain_M._Banks.epub, importing now...
Added book ids: 4
[cwa-kindle-epub-fixer] Backing up original file...
[cwa-kindle-epub-fixer] ERROR - Error occurred when backing up cwa-book-ingest/new_1_20250913_100935_164028_The_Player_of_Games_-_Iain_M._Banks.epub to config/processed_b>
Directory does not exist: config/processed_books/fixed_originals/
[cwa-kindle-epub-fixer] Loading provided EPUB...
[...]

Issue #122, opened 2024-11-01
[Feature Request] Instructions for bare-metal/LXC/VM installation without Docker


valantur:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
It would be great to have instructions on how to install CWA without Docker.

**Describe alternatives you've considered**
Clonning the project and running the python app from an LXC container w/o Docker.

**Additional context**
I believe a lot of people who for whatever reason can't or don't want to run Docker would appreciate having an alternative way to run this. I'll try to do it myself in the next couple of days by reading the dockerfile, but some people don't know how to do this and would find it easier to follow step-by-step instructions.

PeteCreation:
Having to learn Docker on top of LXC and Proxmox has been a strain. I've built so many VMs and containers now and still not got this working so a very simple set of instructions would be really helpful.

jmedemar:
Yes, using this in a LXC would be ideal, since I already have calibre-web running in one. A set of instructions to install this into its own container would be appreciated.

Tooelite:
Just to update the issue there is a new script that installs a Proxmox LXC based on the original CWA Docker repo:

https://github.com/vhsdream/calibre-web-automated-lxc

diginfotek:
This would be great. I've tried Tooelite's script but he stopped maintaining it and it currently fails, so no luck there.

I managed to get Calibre-web on an LXC with their manual install instructions, but there is no chance to make it work with CWA. Having the option to run outside Docker would be very much appreciated.

PR #1320, opened 2026-05-01
feat: Rootless container + partial baremetal support

bryce-hoehn:
Hey, I wanted to share something I've been working on the past few days. I have recently migrated to rootless podman where I realized that CWA does not work without root due to its dependency on the lsio base image and the s6 init system. This dependency also means the app cannot function properly outside of a docker container.

I've tried to address both of these issues by moving the subprocess handling out of s6 and into separate bash scripts, which are managed via subprocesses within the Python code. I then rebased the Dockerfile from lsio ubuntu:noble to debian:trixie-slim. This change also simplified the build steps significantly as debian-trixie now has Python 3.13 and lsof 4.99.4. Finally, I replaced the abc user (911:911) with a rootless calibre user (1000:1000).

This would however be a **breaking change** which will have to be fixed by manually changing the file ownership. It would also drop support for the `PUID` and `PGID` environmental variables, but the app would be defaulting to 1000:1000 anyway. For users using an ID other than 1000:1000, this could cause a permission error, but it would be easily fixed by adding `user: 1001:1001:` into the docker-compose.yml.

Might have to add some sort of migration step or warning.

This is my first major contribution of this scale to an open source project so please let me know if there's any changes I should make or if I did something wrong!

I have not done much testing yet. I can confirm that the web-ui loads and is at least able to open books. The EPUB Fixer service also seems to be working. The Dockerfile should be in good shape, but I'm much less familiar with multi process management and the CWA background scripts.

### Dockerfile
- Base image changed from lsio base image to debian:trixie-slim
- Removed deadsnakes PPA. Debian Trixie now ships Python 3.13 natively
- Removed LSOF build step. Debian Trixie now ships 4.99.4, which has the bugfix
- Replaced abc user with calibre user. UID and GID changed from 911 to 1000
- Removed several unused runtime dependencies from the build step
- Renamed virtual environment from /lsiopy to /venv

### Process Management (replaces s6-overlay)
- New `cps/services/background_services.py` — starts and manages all background services as subprocesses
- Graceful shutdown: SIGTERM → 5s timeout → SIGKILL fallback
- `atexit` safety net for non-signal exit paths
- Auto-restart loops in `cwa-ingest-service.sh` and `metadata-change-detector.sh`

### Service Scripts
- Moved from `root/etc/s6-overlay/s6-rc.d/*/run` to `scripts/services/*.sh`
- All services are now standard bash scripts, initiated within the Python code instead of s6. This should also partially address https://github.com/crocodilestick/Calibre-Web-Automated/issues/122, https://github.com/crocodilestick/Calibre-Web-Automated/issues/185, and https://github.com/crocodilestick/Calibre-Web-Automated/issues/639


Thovi98:
This would be such an enhancement! Thanks a lot!

PR #1327, opened 2026-05-02
Fix book delete CSRF and redirect handling

okms:
## Summary
- fix delete confirmation redirect handling so the saved return location is passed to the generated POST form instead of assigning window.location
- add explicit X-CSRFToken headers to book table JSON POST actions, including bulk delete and related bulk edit/archive/read actions

## Root cause
The detail-page delete flow called postButton with location=loc, which assigns the browser's global location object in JavaScript. Some table bulk actions also relied on global AJAX CSRF setup and could reach Flask-WTF without a CSRF token header.

## Validation
- git diff --cached --check
- node --check cps/static/js/main.js
- node --check cps/static/js/table.js

### DIFF
--- a/cps/static/js/main.js
+++ b/cps/static/js/main.js
@@ -20,7 +20,7 @@ function getPath() {
     return jsFileLocation.substr(0, jsFileLocation.search("/static/js/libs/jquery.min.js"));  // the js folder path
 }
 
-function postButton(event, action, location=""){
+function postButton(event, action, redirectLocation=""){
     event.preventDefault();
     var newForm = jQuery('<form>', {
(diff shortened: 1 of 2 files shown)


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.25 at [`8ec4c58`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/8ec4c58). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest`. Thanks @okms.

PR #1330, opened 2026-05-04
Fix metadata modal crash when provider results have missing fields

I-Would-Like-To-Report-A-Bug-Please:
Metadata search could successfully return results from a provider, but the edit-book metadata modal failed to render them if optional fields were missing or null.

This happened with lubimyczytac.pl: the backend returned valid results for Tajemnica tajemnic, but those records had tags: null. The frontend template assumed book.tags was always an array and called book.tags.length, causing a browser-side render error. From the user’s perspective, metadata search appeared to return no results even though the backend response was valid.

Similar provider contract issues existed in Amazon providers, where authors could be set to an empty string even though the UI expects an array.

Fix.
This PR makes provider output and frontend rendering more robust:

- lubimyczytac now returns an empty list for missing tags instead of None.
- Amazon and AmazonJP now use an empty authors list instead of an empty string.
- The metadata modal now defensively handles missing/null authors, tags, and identifiers.
- Added a regression test covering the no-tags lubimyczytac case.

### DIFF
--- a/cps/metadata_provider/amazon.py
+++ b/cps/metadata_provider/amazon.py
@@ -68,7 +68,7 @@ class Amazon(Metadata):
             try:
                 match = MetaRecord(
                     title = "",
-                    authors = "",
+                    authors = [],
                     source=MetaSourceInfo(
                         id=self.__id__,
                         description="Amazon Books",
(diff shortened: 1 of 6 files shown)


new-usemame:
Backported into the community CWA-derived build [Calibre-Web-NextGen v4.0.104](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.0.104) — `docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.104` (or `:latest`). All 5 file changes plus 8 source-pin regression tests. Thanks @I-Would-Like-To-Report-A-Bug-Please for the diagnosis + fix. For users hitting the empty modal: pulling the image above is enough; metadata search now renders results from lubimyczytac/amazon/amazonjp even with missing optional fields.

PR #1331, opened 2026-05-04
Fix fetched metadata not persisting on first save after title rename

I-Would-Like-To-Report-A-Bug-Please:
**Problem:**

When applying fetched metadata that changed the book title or author, the first save also renamed the book directory. The existing flow staged fetched metadata first, then ran the
directory update.

In this path the UI could report a successful save, the cover could update, and metadata enforcement could run, but fields such as description/tags were not persisted to `metadata.db`.
Saving the same metadata a second time worked because the title/author rename was no longer needed.

Moving the directory update ahead of the remaining metadata staging fixes the first-save path while keeping a single directory update per save.

**Fix:**

Moves the title/author directory update earlier in the edit-book save flow, before fetched metadata fields like comments and tags are staged.

Adds a regression test to ensure the directory update continues to run before comments/tags are applied.

### DIFF
--- a/cps/editbooks.py
+++ b/cps/editbooks.py
@@ -858,7 +858,7 @@ def do_edit_book(book_id, upload_formats=None):
         title_author_error = None
         input_authors = [author.name for author in book.authors]
 
-        # Stage 1: Collect all metadata changes and apply them to the book object in the session
+        # Stage 1: Apply title/author changes before touching the remaining metadata fields.
         if "title" in to_save:
             title_change = handle_title_on_edit(book, to_save["title"])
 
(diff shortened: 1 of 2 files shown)


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.25 at [`3129d90`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/3129d90). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest`. Thanks @I-Would-Like-To-Report-A-Bug-Please.

PR #1333, opened 2026-05-04
fix: update LubimyCzytac metadata parsing

I-Would-Like-To-Report-A-Bug-Please:
This should update lubimyczytac.pl provider, which was slightly outdated(missing description & publisher) .

Adds tests.


I-Would-Like-To-Report-A-Bug-Please:
@new-usemame - any chance this could go into nextgen version too?

PR #1335, opened 2026-05-05
fix: bound healthcheck curls and close gevent keepalives

I-Would-Like-To-Report-A-Bug-Please:
When the app stops responding, Docker's shell-form healthcheck can time out while leaving the child curl process behind. Repeated failures accumulate stuck curls, and reverse-proxy keepalive sockets can remain attached to the gevent process after the client side is gone.

Give curl its own connect and total request timeouts so healthcheck children exit before Docker kills the shell, and force gevent request handlers to close HTTP connections after each response. This avoids stale keepalive sockets and makes healthcheck failures bounded.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -299,4 +299,4 @@ VOLUME /calibre-library
 # Health check for container orchestration
 # Uses shell form to support environment variable substitution for CWA_PORT_OVERRIDE
 HEALTHCHECK --interval=30s --timeout=3s --start-period=120s --retries=3 \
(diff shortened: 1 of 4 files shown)


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.86 at [`92c8cfa`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/92c8cfa). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest`. Thanks @I-Would-Like-To-Report-A-Bug-Please.

PR #1336, opened 2026-05-05
fix: sync applied metadata descriptions before save

I-Would-Like-To-Report-A-Bug-Please:
Metadata search results can update the TinyMCE comments editor without updating the backing textarea that the edit form submits. That makes the fetched description visible in the UI while the first save can still post the previous or empty comments value.

Normalize missing provider descriptions to an empty string, save TinyMCE after applying the fetched description, and rename the result button from "Save" to "Apply" so it reflects that the action only copies metadata into the edit form.

### DIFF
--- a/cps/static/js/get_meta.js
+++ b/cps/static/js/get_meta.js
@@ -106,10 +106,12 @@ $(function () {
       )
     );
     if (updateItems.description) {
+      var description = book.description || "";
       if (typeof tinymce !== "undefined" && tinymce.get("comments")) {
-        tinymce.get("comments").setContent(book.description);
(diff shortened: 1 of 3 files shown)


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.26 at [`5d6943d`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/5d6943d). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest`. Thanks @I-Would-Like-To-Report-A-Bug-Please.

PR #1337, opened 2026-05-05
fix: reduce metadata background save races

I-Would-Like-To-Report-A-Bug-Please:
Metadata saves can kick off noisy duplicate-status polling and the metadata enforcer can miss supported files immediately after Calibre renames a book folder. In practice this makes metadata edits look stuck or intermittently fail post-save enforcement even though the supported EPUB/AZW3 file exists shortly afterward.

Stop duplicate notifications from continuously polling stale/pending cache states on normal page loads, and make the cover/metadata enforcer retry supported file discovery when Calibre's DB says a supported format should be present. Add diagnostic output listing visible files when the expected format never appears.

### DIFF
--- a/cps/static/js/duplicate-notifier.js
+++ b/cps/static/js/duplicate-notifier.js
@@ -183,18 +183,9 @@
             }
         }
 
-        if ((data.needs_scan || data.stale) && !isModalActive()) {
-            startStatusPolling();
-            return;
-        }
-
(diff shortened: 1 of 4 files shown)


I-Would-Like-To-Report-A-Bug-Please:
@new-usemame - any chance this could go into nextgen version too?

PR #1343, opened 2026-05-09
Fix minor compatibility issues with Kobo's Android app

shavitmichael:
The Kobo android app uses the same sync protocol as Kobo devices. For a few small cases it has stricter expectations compared to Kobo devices however (e.g. series field must be set in metadata, even if empty).

This is mostly useful for testing changes to calibre-web's sync implementation against an official Kobo sync client without access to a Kobo device (e.g. by running the Kobo android app in an emulator)

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -598,7 +598,7 @@ def get_metadata(book):
                         "Url": get_download_url_for_book(book.id, book_data.format),
                         # The Kobo forma accepts platforms: (Generic, Android)
                         "Platform": "Generic",
-                        # "DrmType": "None", # Not required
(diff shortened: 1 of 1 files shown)


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.72 at [`dd94fb1`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/dd94fb1). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest`. Thanks @shavitmichael.

PR #1344, opened 2026-05-10
Low hanging optimizations of the Kobo Sync request

shavitmichael:
This pull request is part of a larger rewrite of the Kobo sync request handler. Specifically, this defers or otherwise avoids some of the unnecessary side-effects comitted by the sync request handler (such as book conversion, default database entry creation, etc).

Over-all, these save ~15 seconds on a *single* request to the Sync endpoint. On the first request from a new device, the request can be paginated into multiple rounds, in which case these savings apply to every round.

### DIFF
--- a/cps/helper.py
+++ b/cps/helper.py
@@ -5,6 +5,7 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 # See CONTRIBUTORS for full list of authors.
 
+import glob
 import os
 import random
 import io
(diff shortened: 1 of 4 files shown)


shavitmichael:
Latest force-push: fixed issue where get_or_create_reading_state was incorrectly and unecessarily refactored in an earlier commit in the chain. This caused exceptions on calls to the state handler

new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.146 at [`27b334c`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/27b334c). Drop-in image for users hitting this: `ghcr.io/new-usemame/calibre-web-nextgen:latest`. Thanks @shavitmichael.

Issue #1007, opened 2026-02-01
[Feature Request] Multi-device Kobo sync with same token: new books delivery requires manual forcing and can wipe downloads/progress + DB constraint blocks sharing Hardcover token


jaimetur:
Summary

Using the same Kobo sync token across multiple Kobo devices (same account) is partially supported: reading progress can sync.
However, newly added books are not reliably delivered to all devices, and forcing sync/push can cause previously downloaded books and reading progress to be lost.
Additionally, the current database schema prevents using the same Hardcover token across multiple user accounts because user.hardcover_token is UNIQUE.

Current behavior

Multiple Kobo devices can be linked using the same Kobo sync token, but:

Only the first device that uses the token automatically pulls newly added books.

Other devices require a forced sync and then another pull to get the new books.

With N devices, this must be repeated N times for every book addition.

The most frustrating part: after forcing a push/sync, previously downloaded books can disappear, and reading progress sync can be lost.

Expected behavior

If multiple Kobo devices use the same Kobo sync token (same user/account):

Newly added books should be delivered to all linked devices without manual forcing per device.

Forced sync/push should not wipe existing downloads or reading progress.

Database limitation (Hardcover token)

Currently it is not possible to use the same Hardcover token across two different accounts because the DB schema enforces uniqueness:

user.hardcover_token has a UNIQUE constraint (autoindex)

This prevents assigning the same Hardcover token to multiple users.

Steps to reproduce

Link Device A, Device B (and optionally Device C) using the same Kobo sync token.

Add a new book to the library.

Observe:

Device A pulls the new book automatically.

Device B/C do not receive it unless you force sync and then pull manually.

Force push/sync on Device B/C.

Observe:

Previously downloaded books may disappear.

Reading progress sync may be lost/reset.

Impact

Highly frustrating workflow for users with 2+ Kobo devices.

Scaling problem: more devices = more manual sync operations for each new book.

Risk of data loss (downloads/progress), which discourages using multi-device setups.

Possible solutions / proposals
Option A (DB-level): allow shared Hardcover token across users

Remove the UNIQUE constraint on user.hardcover_token so the same Hardcover token can be used by multiple users.

This could make it easier to model multiple devices/users under one Hardcover identity (depending on the architecture).

Option B (preferred product behavior): allow multiple Kobo sync tokens per user

Store multiple Kobo sync tokens under the same user account.

Then deliver/push new books across all tokens/devices linked to that user, without per-device manual forcing.

[...]

Issue #323, opened 2025-03-22
[bug] Kobo sync: doesn't pick up conversion to compatible format


itsthejb:
**Describe the bug/problem** 

When using the Kobo sync feature; given you have an entry in Calibre web in an incompatible format, e.g. PDF. Calibre won't pick it up if you add a compatible format later. Instead it's necessary to do a full resync in order for it to display

**To Reproduce**

1. Add PDF file to library
2. Sync with Kobo
3. Note that the new entry doesn't appear on the Kobo device
4. Convert the PDF to EPUB in Calibre
5. Sync again
6. The entry will still not be synced

**Expected behavior**

The new entry should sync to Kobo as soon as it has a compatible format. Perhaps CW needs to delete sync status for an entry when a format is changed?

Issue #216, opened 2025-01-31
[Feature Request] Add ability to resend books in next kobo sync


Waggleworx:
**Is your feature request related to a problem? Please describe.**
When a kobo sync fails, sometimes CWA thinks books were sent successfully, when they were not.  Corresponding rows are added to kobo_synced_books, causing the book to never be resent unless forcing a full kobo resync.  I think a fair number of us likely run CWA on synology NAS's, and the performance constraints combined with kobo device's hardcoded 30 second timeout cause this issue to rear its ugly head.

**Describe the solution you'd like**
I would like Book List to be enhanced with a button that puts books in a condition to send in the next kobo sync.  If a user isn't using sync shelves, I believe this involves just deleting the corresponding rows from kobo_synced_books.  If a user is modifying sync shelves it will require either a logic change to create an override kobo sync queue or I think change book_shelf_link.date_added to a datetime after the last kobo sync.

**Describe alternatives you've considered**
Currently my workaround is to stop the container and remove corresponding rows from kobo_synced_books.  If using a sync shelf, the books also must be removed/readded to the sync shelf to make it into the next sync.  This workflow is awful since there's no way to bulk remove books from a shelf and it requires stopping the container to manually delete rows in app.db (ew).

**Additional context**

Waggleworx:
I don't know this project's stance on community contribution, but I'm hoping to begin work on this PR myself in the near future when time allows.

Waggleworx:
A tangent additional Book List enhancement that would help greatly is a "Select all results" button in Book List to facilitate bulk changes in combination with the search utility.  I wish the search was more feature-rich, but I can workaround that for the most part if I could only select all results

Domoel:
If I understands you correctly, we need to have a multi select option while using Calibre web to be able to select the desired books for a future sync. Or how should that work?

new-usemame:
Per-book resend shipped in v4.0.73 of [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen), a community-maintained CWA build. **Admin → Edit User → "Resend one book to this Kobo"**: enter the book ID, click Resend, and the device receives the book on its next sync (clears the per-user sync row + bumps the book's `last_modified` so it re-delivers as a fresh entitlement).

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.73
```

(or `:latest`)

[...]

PR #1345, opened 2026-05-10
Redesign the Kobo book syncing flow (multi-device support, 100-250ms per sync request instead of 10-30seconds, etc)

shavitmichael:
The previous implementation relied on server-side tracking of each sync request which suffered from the following issues:
- Slow sync responses because of database changes on each request
- Race conditions between a paginated sync response and concurrent server side change
- Race conditions on a dropped sync response (book considered synced by server but not client)
- Multiple devices can't use the sync endpoint
- Difficulty distinguishing metadata changes from other changes

The new design changes the pagination method to improve robustness, and eliminates (nearly) all server side side-effects on a sync request.
In testing, this implementation can sync a 5000 book library from scratch in 45 seconds with all books in kobo shelves. Syncing the same library from scratch again but this time with 0 books in kobo syncable shelves took 20 seconds. For comparison, I was seeing latencies of up to 30 seconds just for syncing the first 100 books of this library before this change.
This implementation also let's multiple devices sync to the same kobo endpoint, and enables client-side reset of the SyncToken to trigger a full resync.

Note: This pull request contains  #1344 and #1343 in its history.

High level design changes:

**1. A new KoboBookVisibility table is added to track which books are visible to kobo**
 - Kobo shelf membership changes update the KoboBookVisibility table
 - The first Sync before pagination re-computes magic shelf memberships and updates the KoboBookVisibility table based on its result.

**2. The sync handler now makes a single database query to build NewEntitlement, ChangedEntitlement and ChangedProductMetadata responses:**
 - A ChangedEntitlement entry is emitted whenever a book becomes visible or invisible (i.e removed)
 - A NewEntitlement entry is emitted whenever a new book is added (vis-a-vis the device SyncToken). If the book shouldn't be visible, a blank BookMetadata is included so that the device can process a later ChangedEntitlement entry that makes the book visible (devices won't accept a ChangedEntitlement entry if no entitlement existed before). This allows us to remove the server-side tracking of which books have been synced to the device. Instead we only need track visibility changes using the SyncToken's book_visibility_last_modified field
 - A ChangedProductMetadata entry is emitted whenever a book metadata is changed, or whenever a book becomes visible (to override the previously blank metadata we may have sent on a non-visible NewEntitlement)

**3. Pagination of large sync responses is rewritten for robustness against race conditions:**
 - A snapshot timestamp is recorded at the beginning of a multi-round response and recorded in a new SyncTokenPagination object in the SyncToken.
 - All queries accross paged-sync requests are filtered to entries with modifications below that snapshot timestamp.
 - Once we've gotten through all books in the table, we can update the tokens last_created/last_modified/etc timestamps to the snapshoted timestamp and clear out the SyncTokenPagination object.
 - If any modifications to the library were concurrently made server side during these exchanges, they will be picked up on the next request after we cleared SyncTokenPagination.

**For backwards compatibility:**
 - "Force full sync" is now implemented as a kobo_force_full_sync boolean on the User model. Setting it causes the next sync request to discard the device's incoming token and replay the library; the flag is cleared once that sync starts. This isn't really necessary anymore, since Kobo devices have a setting to "repair" the sync state which resets the SyncToken on the client side.
 - The kobo_synced_books and kobo_book_sync_state tables are dropped on startup.
 - The SyncToken version number is incremented, and a full sync is forced to ensure no compatibility issues with the previous state.

**Out of scope:**
 - Syncing of tag/collection membership to the device is kept out of scope of this pull request since it's already grown quite large. But note that it is bug prone and could use a similar effort. The only extent to which it is touched here is ensuring that shelf membership correctly affects book visibility. Oh and tag memberships are only computed and returned on the final request of a pagination exchange instead of every request.

Related issues:
https://github.com/janeczku/calibre-web/issues/1276
https://github.com/janeczku/calibre-web/issues/3331
https://github.com/crocodilestick/Calibre-Web-Automated/issues/1112

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -787,7 +787,12 @@ def edit_list_user(param):
                 elif param == 'email':
                     user.email = check_email(vals['value'])
                 elif param == 'kobo_only_shelves_sync':
+                    old_val = user.kobo_only_shelves_sync
                     user.kobo_only_shelves_sync = int(vals['value'] == 'true')
+                    if old_val == 0 and user.kobo_only_shelves_sync == 1:
+                        kobo_sync_status.update_on_sync_shelfs(user.id)
+                    elif old_val == 1 and user.kobo_only_shelves_sync == 0:
+                        kobo_sync_status.on_sync_shelves_disabled(user.id)
                 elif param == 'kindle_mail':
                     user.kindle_mail = valid_email(vals['value']) if vals['value'] else ""
                 elif param == 'kindle_mail_subject':
@@ -1248,8 +1253,14 @@ def ajax_pathchooser():
 
 
 def do_full_kobo_sync(userid):
-    count = ub.session.query(ub.KoboSyncedBooks).filter(userid == ub.KoboSyncedBooks.user_id).delete()
-    message = _("{} sync entries deleted").format(count)
+    # Flag this user so the next Kobo sync request ignores its incoming SyncToken
+    # and replays the full library. The flag is cleared when that sync starts.
+    user = ub.session.query(ub.User).filter(ub.User.id == userid).one_or_none()
+    if user is None:
+        message = _("User not found")
+        return Response(json.dumps([{"type": "danger", "message": message}]), mimetype='application/json')
+    user.kobo_force_full_sync = True
+    message = _("Full Kobo sync requested; the next sync from this user's device will replay the library.")
     ub.session_commit(message)
     return Response(json.dumps([{"type": "success", "message": message}]), mimetype='application/json')
 
@@ -2262,7 +2273,6 @@ def _db_configuration_update_helper():
             ub.session.query(ub.Bookmark).delete()
             ub.session.query(ub.KoboReadingState).delete()
             ub.session.query(ub.KoboStatistics).delete()
-            ub.session.query(ub.KoboSyncedBooks).delete()
             helper.delete_thumbnail_cache()
             ub.session_commit()
             # deleted visibilities based on custom column and tags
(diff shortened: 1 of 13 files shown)


shavitmichael:
More related issues:  #1007 , #216 , #323, [#3298 ](https://github.com/janeczku/calibre-web/issues/3298)

shavitmichael:
Looks like there's an issue in one of the latter commits in this chain. Although we're sending back a SyncToken with kobo_store_pagination=false on the last response, the next request from the device has kobo_store_pagination set back to true. I suspect the device maybe isn't be updating its stored SyncToken if the response is empty (or if x-kobo-sync isn't "continue")

shavitmichael:
Ok confirmed that this is the case on the Kobo Android app: it specifically checks whether the Sync results contains any items before savings the SyncToken. For now I'll simply drop the commit in question [48609c1](https://github.com/crocodilestick/Calibre-Web-Automated/pull/1345/commits/48609c10a63eb4c4d8d7797c3df05649665afa87) (and technically [e534103](https://github.com/crocodilestick/Calibre-Web-Automated/pull/1345/commits/e534103d1800ace1259d3047585ca9def0ec4a63) as well). Concretly this is a minor setback for this pull request: it just means that we may in some rare cases make an unecessary query against the calibre-web database when we are proxying the request to the kobo store.

FWIW, there is a hacky solution to keep those commits: we could always return some item in the response (e.g. a DeletedTag with an invalid id) just to get the SyncToken save. But it's not clear whether this is even an issue on real Kobo devices, and IMO it's not really worth it.

raphi011:
Out of curiosity, is this going to land any time soon? I'm currently running a custom image with this PR to make kobo sync work properly.

shavitmichael:
This repo hasn't landed any pull requests since Frebruary-ish. There's been some discussion about trying to get it into the next-gen fork here, but it's still a work in progress: https://github.com/new-usemame/Calibre-Web-NextGen/issues/347#issuecomment-4589460778 .

Issue #1256, opened 2026-04-04
[bug] Critical Bug: Web UI 504 Timeout and Database Locking during Ingest


kanjieater:
The Web UI becomes unresponsive (returning 504 Gateway Timeouts or 502 Bad Gateway) and the ingestion process slows to a crawl or crashes when adding books to a library. The background ingest-processor locks the SQLite database while performing metadata fetching and EPUB fixing, preventing the gevent web server from responding to client requests.

To Reproduce

Bind a existing Calibre library  to the container.

Drop multiple books (especially PDFs or books requiring metadata) into the /cwa-book-ingest folder.

Attempt to navigate the Web UI while the logs show [ingest-processor] is active.

Observe "Read timed out" warnings in logs and 504 errors in the browser.

Expected behavior
The ingestion process should ideally run as a low-priority background task that does not lock the database long enough to cause the Web UI to time out, or it should provide a queue system that prevents resource exhaustion on NAS/low-power hardware.

Screenshots
(N/A - Log-based report)

Configuration (please complete the following information):

OS: Me: Ubuntu, others: (Synology DSM / Docker / Portainer)

Hardware: Me: Custom Linux box, others: Synology NAS (DS224+ / x86_64)

Version: v4.0.6

Additional context
[Discussion here](https://github.com/crocodilestick/Calibre-Web-Automated/discussions/1076)

kanieldeith:
Just to add to this, experiencing very similar behaviour but rather than ingesting books, triggered this condition when fetching metadata to update existing books. Wonder if a similar pattern occurs during ingest when retrieving/setting metadata. Seems to be a Python GIL/sqlite deadlock as the container logs stop entirely. 

Disclaimer: Did use Claude to assist with formatting my notes into the below but debugging was done manually by me. 

---

When saving book metadata after fetching a cover from a metadata provider (e.g. Hardcover), the Calibre-Web process freezes entirely and becomes unresponsive. The health check fails, the container is marked unhealthy, and a restart is required. The issue is intermittent — it typically occurs after several rapid successive saves.

**Root cause (identified via py-spy)**

A GIL + SQLite deadlock between two threads:

- Thread 1 (main request): executing edit_book_tags → fetchall on a SQLite query that invokes Python UDF callbacks (e.g. title_sort) — requires the GIL to execute the callbacks
[...]

Issue #1217, opened 2026-03-21
[bug] Several DB operation fails because HTTPConnectionPool(host='127.0.0.1', port=8083): Read timed out. (read timeout=5)


joselsegura:
**Describe the bug**
While CWA is importing ebooks from the ingest directory, after a while the web becomes inaccessible and those logs are shown in the docker logs output of the container. Apparently the ingest process continues, but I cannot be sure if they are being added to the database or not.

**To Reproduce**
1. Add an insane number of ebooks in the ingest directory
2. Wait to the ingest process to start
3. Wait for a few minutes
4. See errors like the followingÑ

```
WARN: Failed to call DB refresh endpoint: HTTPConnectionPool(host='127.0.0.1', port=8083): Read timed out. (read timeout=5)
Continuing despite session refresh failure - books may require manual refresh
WARN: Failed to invalidate duplicate cache: HTTPConnectionPool(host='127.0.0.1', port=8083): Read timed out. (read timeout=5)
WARN: Failed to schedule duplicate scan via web API: HTTPConnectionPool(host='127.0.0.1', port=8083): Read timed out. (read timeout=5)
```

**Expected behavior**
- No errors in the logs.
- Web responsible during the ingestion process.

**Screenshots**
N/A

**Configuration(please complete the following information):**
 - OS: Debian, Docker Engine Community 29.0.2
 - Hardware: x86_64 server
 - Version v4.0.6

**Additional context**

Datinhkhach:
I encountered the same error.

BarBaar44:
Same here.. causes a 404 on the frontend..

Some more logging:

> [ingest-processor] No users with auto-send enabled found
[ingest-processor] Refreshing Calibre-Web database session...
[ingest-processor] WARN: Failed to call DB refresh endpoint: HTTPConnectionPool(host='127.0.0.1', port=8083): Max retries exceeded with url: /cwa-internal/reconnect-db (Caused by ConnectTimeoutError(<HTTPConnection(host='127.0.0.1', port=8083) at 0x7fba94b15bf0>, 'Connection to 127.0.0.1 timed out. (connect timeout=5)'))
[ingest-processor] Continuing despite session refresh failure - books may require manual refresh
[...]

Issue #1082, opened 2026-02-08
[bug] Failing to import ebook using folder ingestion - database is locked


rmcbarreto:
**Describe the bug**
I have done a fresh install of Calibre-Web-Automated using docker compose.

I was importing an ebook, but it failed and threw this error:
```
[cwa-ingest-service] New file detected - /cwa-book-ingest/George R.R. Martin - A Dança Dos Dragões.epub - Starting Ingest Processor...
[cwa-ingest-service] Configured timeout: 900s, Safety timeout: 2700s
[cwa-kindle-epub-fixer] NETWORK_SHARE_MODE=true detected; skipping chown of /config/epub-fixer.log
[2026-02-08 16:48:29,425]  INFO {cps:92} ProxyFix configured to trust 1 proxy(ies) for X-Forwarded-* headers
[ingest-processor] GDrive functionality available
[ingest-processor] Auto-send and metadata functionality available
[2026-02-08 16:48:30,715]  WARN {cps.db:789} WAL mode disabled for calibre/app_settings (NETWORK_SHARE_MODE=true)
[ingest-processor] Lock acquired successfully (PID: 1500)
[ingest-processor] Checking if file is ready (timeout: 15 minutes): George R.R. Martin - A Dança Dos Dragões.epub

[ingest-processor]: No conversion needed for George R.R. Martin - A Dança Dos Dragões.epub, importing now...
[cwa-kindle-epub-fixer] Backing up original file...
[cwa-kindle-epub-fixer] Loading provided EPUB...
[cwa-kindle-epub-fixer] Checking linking to body ID to prevent unresolved hyperlinks...
[cwa-kindle-epub-fixer] Checking META-INF/container.xml integrity...
[cwa-kindle-epub-fixer] Checking UTF-8 encoding declaration...
[cwa-kindle-epub-fixer] Checking language field tag is valid...
[cwa-kindle-epub-fixer] Checking for stray images...
[cwa-kindle-epub-fixer] 4 issues fixed with /cwa-book-ingest/George R.R. Martin - A Dança Dos Dragões.epub:
   01 - Converted META-INF/container.xml from ascii to utf-8
   02 - Converted page_styles.css from ascii to utf-8
   03 - Converted stylesheet.css from ascii to utf-8
   04 - Added utf-8 to XML declaration missing encoding info in META-INF/container.xml
[cwa-kindle-epub-fixer] Writing EPUB...
[...]

PR #1353, opened 2026-05-14
Improve duplicate scanning performance

navels:
## Summary

Improved duplicate scanning performance and reliability by replacing repeated broad duplicate checks with an indexed duplicate cache, incremental updates, and safer coordination with active imports.

## Changes

* Add a duplicate index for normalized duplicate keys so duplicate groups can be rebuilt from indexed rows instead of repeatedly scanning the full library.
* Require a one-time full duplicate scan after upgrade to establish the initial index baseline, using the existing Duplicates page as the user-facing entry point.
* Update duplicate state incrementally after successful imports, metadata edits, and deletions.
* Keep the Duplicates badge, Duplicates page status, cleanup results, and indexed cache in sync after scans and auto-resolution.
* Prevent full duplicate scans from racing active ingest: manual scans are blocked during import, scheduled scans skip while import is active, and ingest waits if a full scan is already running.
* Replace alert-based duplicate cleanup feedback with Bootstrap-style modal feedback.
* Add unit coverage for index rebuilds, incremental scan rewiring, queue settings, delete maintenance, ingest coordination, and cache consistency.

## Testing

* Added unit tests for duplicate index rebuilds, incremental duplicate scans, delete/index maintenance, queue settings, and ingest coordination.

* Manually tested Docker import flows with duplicate EPUBs, full duplicate scans, duplicate badge updates, duplicate page updates, and auto-resolution cleanup behavior.

## Related Issues

Likely helps with:

- #1256
- #1217
- #1082

Disclaimer: I used Codex (gpt-5.5) and Claude (claude-sonnet-4-6) for various parts of this work and I reviewed it myself before submitting this PR.


new-usemame:
Backported into Calibre-Web-NextGen [v4.0.76](https://github.com/new-usemame/Calibre-Web-NextGen/releases/tag/v4.0.76) (`3047debf`). Thanks @navels — the incremental index design lands cleanly.

For end-users hitting slow duplicate scans on large libraries, the community-maintained CWA build has it now:

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:v4.0.76
```

(or `:latest`)

After upgrade, opening **Duplicates** shows a one-time "Run Full Duplicate Scan" notice — click once to build the baseline. Imports then update the index in milliseconds instead of rescanning.

[Tracker on the fork](https://github.com/new-usemame/Calibre-Web-NextGen/issues) for follow-ups.

crocodilestick:
@navels — great PR, really solid work on the duplicate index architecture. I had a couple of small suggestions after reviewing:

1. Added has_active_task_of_type() to WorkerThread (worker.py)

There are two places that iterate WorkerThread.tasks to check for an active duplicate scan: _duplicate_full_scan_running() in cwa_functions.py and the inline loop in _duplicate_scan_transiently_pending() in duplicates.py. I extracted the pattern into a reusable method on WorkerThread:

It accepts an optional extra_check lambda for additional filtering (used by the full-scan check for full_scan=True. This keeps the terminal-stats logic and iteration in one place and makes the intent clearer at the call sites.

Also updated _duplicate_full_scan_running() to delegate to it — went from 10 inline lines to a clean 4-line call.

2. Moved dismiss notice files from app to config

The duplicate index setup notice dismiss files were being written to /app/cwa_duplicate_index_setup_notice_*. That works across container restarts, but not across image rebuilds (unlike every other persistent state file the project uses — ingest status, logs, profiles, etc. — which all live under config. Moved the three references in render_template.py and duplicates.py to use config instead.

Issue #234, opened 2025-02-04
[bug] Epub-fixer doesn't respect PUID/PGID settings


blinky-z:
**Describe the bug**
It seems Epub-fixer runs with a default user (root) and doesn't use the PUID/PGID env vars. The Calibre itself does use them.

This issue is minor though. It doesn't affect the ingestion process. There's only one exception in the log:
```
Traceback (most recent call last):
  File "/app/calibre-web-automated/scripts/ingest_processor.py", line 12, in <module>
    from kindle_epub_fixer import EPUBFixer
  File "/app/calibre-web-automated/scripts/kindle_epub_fixer.py", line 31, in <module>
    file_handler = logging.FileHandler('/config/epub-fixer.log', mode='w')
  File "/usr/lib/python3.10/logging/__init__.py", line 1169, in __init__
    StreamHandler.__init__(self, self._open())
  File "/usr/lib/python3.10/logging/__init__.py", line 1201, in _open
    return open_func(self.baseFilename, self.mode,
PermissionError: [Errno 13] Permission denied: '/config/epub-fixer.log'
```

This file `epub-fixer.log` is created by epub-fixer with the root user as an owner and is not accessible by Calibre due to permission issues.

**To Reproduce**
1. Import any epub book

**Expected behavior**
There's no exception in the logs.

**Configuration(please complete the following information):**
 - OS: Ubuntu 22 + Kubernetes
 - Hardware: x86 server

crocodilestick:
Thanks for the report! Fix coming later today

crocodilestick:
Fixed in V3.0.3

dgorley:
I'm still experiencing this problem in v3.0.4.

blinky-z:
Can you confirm your epub folder and the files inside are owned by the same user running the container?

dgorley:
I thought so but it was human error on my end.  All good, disregard, and thanks for the quick reply!

Issue #108, opened 2024-10-12
[bug] HTTP Error 500 after changing settings and restarting


Wit4r7:
**Describe the bug**
After enabling the upload function by user in the basic configuratio, turning on the Thumbnail creation, uploading and fetching some metdata and restarting the container, I got an HTTP 500 Internal server error. 

```
Calibre-Web
500 Internal Server Error
The server encountered an internal error and was unable to complete your request. There is an error in the application.
Traceback (most recent call last):
File "/lsiopy/lib/python3.10/site-packages/flask/app.py", line 2190, in wsgi_app
response = self.full_dispatch_request()
File "/lsiopy/lib/python3.10/site-packages/flask/app.py", line 1486, in full_dispatch_request
rv = self.handle_user_exception(e)
File "/lsiopy/lib/python3.10/site-packages/flask/app.py", line 1484, in full_dispatch_request
rv = self.dispatch_request()
File "/lsiopy/lib/python3.10/site-packages/flask/app.py", line 1469, in dispatch_request
return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
File "/app/calibre-web/cps/usermanagement.py", line 35, in decorated_view
return login_required(func)(*args, **kwargs)
File "/lsiopy/lib/python3.10/site-packages/flask_login/utils.py", line 290, in decorated_view
return current_app.ensure_sync(func)(*args, **kwargs)
File "/app/calibre-web/cps/web.py", line 806, in index
return render_books_list("newest", sort_param, 1, page)
File "/app/calibre-web/cps/web.py", line 417, in render_books_list
return render_title_template('index.html', random=random, entries=entries, pagination=pagination,
File "/app/calibre-web/cps/render_template.py", line 153, in render_title_template
cwa_update_notification()
File "/app/calibre-web/cps/render_template.py", line 141, in cwa_update_notification
update_available, current_verision, tag_name = cwa_update_available()
File "/app/calibre-web/cps/render_template.py", line 119, in cwa_update_available
tag_name = response.json()['tag_name']
KeyError: 'tag_name'
```

**To Reproduce**
Not sure. As told above. I was changing some settings and updating some covers. I then restarted the container and can't access it now. 

**Expected behavior**
Starting the server :) 

**Screenshots**

**Configuration(please complete the following information):**
Running on a Synology DS720+ with DSM 7.2.1-69057 V5
Image version: 2.0.1

**Additional context**
Calibre-web.log

```
[...]

PR #1359, opened 2026-05-18
Harden author ordering during book list rendering

### LINKED ISSUE #1358: Harden author ordering to avoid UI crashes during ingest
## Summary

A user reported a crash in Calibre-Web-NextGen while importing books and viewing the web UI:

- https://github.com/new-usemame/Calibre-Web-NextGen/issues/234

NextGen appears to have inherited this code path from CWA. CWA currently has the same fragile `cps/db.py::order_authors()` implementation, so this likely belongs upstream here first.

## Observed behavior

The reported stack trace fails while rendering the newest/index page during ingest:

```text
AttributeError: 'NoneType' object has no attribute 'id'
...
cps/web.py -> render_books_list()
cps/db.py -> fill_indexpage()
cps/db.py -> order_authors()
```

The relevant code re-queries `Authors` by `Books.author_sort`, then assumes every returned/fallback author object is valid:

```python
results = self.session.query(Authors).filter(Authors.sort == auth).all()
...
for r in results:
    if r.id in ids:
```

It also appends fallback lookups without checking for `None`.

## Expected behavior

Book-list rendering should tolerate stale, partial, or transiently inconsistent author metadata. A single bad/mismatched author ordering value should not 500 the homepage/newest list during ingest.

## Likely cause

During ingest, `calibredb` mutates `metadata.db` outside the Flask process. If the UI renders a book list while author relationships or `author_sort` values are stale/partial, `order_authors()` can crash instead of falling back to the linked authors.

## Proposed fix

Harden `order_authors()` as a localized display-layer fix:

- Avoid querying `Authors` again by sort while rendering.
- Use the already-linked `book.authors` collection.
- Build dictionaries by author sort/id.
- Skip/log missing sort entries.
- Append remaining linked authors as a fallback.
- Do not append or dereference `None` author objects.

Current Calibre-Web has moved to this general dictionary-based approach in `order_authors()`, but this should be a focused backport rather than pulling in broader search/FTS changes.

Related Calibre-Web context:

- https://github.com/janeczku/calibre-web/pull/3476
- https://github.com/janeczku/calibre-web/pull/3531

The latter has comments cautioning against broad FTS/search backports; this issue is only about the small `order_authors()` hardening.

navels:
## Summary

This hardens book-list author ordering so rendering does not crash when author metadata is stale, partial, or transiently inconsistent during ingest.

The existing `order_authors()` path re-queries `Authors` by `Books.author_sort` while rendering, then assumes every returned/fallback author object is valid. A downstream user reported a crash in this path while importing books in Calibre-Web-NextGen, which appears to inherit the same implementation from CWA.

Fixes #1358.

## Changes

- Orders authors using the already-linked `book.authors` collection instead of issuing extra render-time author queries.
- Handles missing/empty `author_sort`, stale sort values, missing author sort keys, and `None` author entries without raising.
- Falls back to the linked authors when sort reconstruction cannot produce an ordered list.
- Adds unit coverage for normal ordering, stale `author_sort`, `None` author entries, and missing `author_sort`.

## Notes

Current Calibre-Web uses the same general dictionary-based approach for this method. This PR intentionally keeps the change scoped to `order_authors()` and does not backport broader search/FTS changes.

Related downstream report:

- https://github.com/new-usemame/Calibre-Web-NextGen/issues/234

## Testing

```bash
pytest tests/unit/test_order_authors.py -q
```

### DIFF
--- a/cps/db.py
+++ b/cps/db.py
@@ -1084,32 +1084,45 @@ class CalibreDB:
         self.ensure_session()
         for entry in entries:
             if combined:
-                sort_authors = entry.Books.author_sort.split('&')
-                ids = [a.id for a in entry.Books.authors]
-
+                book = entry.Books
             else:
-                sort_authors = entry.author_sort.split('&')
-                ids = [a.id for a in entry.authors]
-            authors_ordered = list()
-            # error = False
+                book = entry
+
+            sort_authors = (book.author_sort or '').split('&')
+            authors_list = [author for author in book.authors if author is not None]
+
+            authors_by_sort = {}
+            authors_by_id = {}
+            for author in authors_list:
+                if author.sort:
+                    authors_by_sort[author.sort] = author
+                if author.id is not None:
+                    authors_by_id[author.id] = author
+
+            authors_ordered = []
+            ids_remaining = set(authors_by_id.keys())
             for auth in sort_authors:
                 auth = strip_whitespaces(auth)
-                # Skip empty author strings to prevent spurious errors
                 if not auth:
                     continue
-                results = self.session.query(Authors).filter(Authors.sort == auth).all()
-                # ToDo: How to handle not found author name
-                if not len(results):
-                    log.error("Author '{}' not found to display name in right order".format(auth))
-                    # error = True
-                    break
-                for r in results:
-                    if r.id in ids:
-                        authors_ordered.append(r)
-                        ids.remove(r.id)
-            for author_id in ids:
-                result = self.session.query(Authors).filter(Authors.id == author_id).first()
-                authors_ordered.append(result)
+
+                author = authors_by_sort.get(auth)
+                if author is None:
+                    log.warning(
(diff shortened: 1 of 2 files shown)


new-usemame:
Backported into [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.85 at [`bc2cb141`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/bc2cb141) — closes our [fork issue #234](https://github.com/new-usemame/Calibre-Web-NextGen/issues/234) (book grid 500ing during concurrent ingest with `CWA_WATCH_MODE=poll`). Preserves the fork-#108 rate-limited drift warning + adds a linked-authors fallback so books with all-stale `author_sort` still render. Users hit by the crash: drop-in `ghcr.io/new-usemame/calibre-web-nextgen:latest`. Thanks @navels — solid diagnosis.

PR #1363, opened 2026-05-19
feat(kobo): track device_id + device_model on bookmarks for multi-device sync

raphi011:
## Summary

Real Kobo devices already send two device-identifying headers on every PUT to `/v1/library/<uuid>/state` that CWA currently discards:

- `x-kobo-deviceid` — opaque per-device identifier (64-char hex on Kobo eReaders; clients like the Kobo iOS app and third-party Kobo-protocol clients may send a UUID instead)
- `x-kobo-devicemodel` — human-readable model name (e.g. `Kobo Libra Colour`)

Capturing both on bookmark writes lets multi-device clients show *who wrote it* with a readable label, not just an opaque ID. Useful for the "you're reading this on Kobo Libra Colour" attribution in cross-device "continue reading" UX, and groundwork for smarter conflict resolution.

## Changes

- **Schema:** two new nullable columns on `KoboBookmark`:
  - `device_id String`
  - `device_model String`
- **Migration:** idempotent `migrate_kobo_bookmark_table` follows the existing `migrate_*_table` pattern in `ub.py` — independent `exists()` probes for each column, so a partial migration from an earlier version of this branch upgrades cleanly.
- **Write path:** `HandleStateRequest` reads both headers on PUT and stores them on the current bookmark. Absence yields `None` and the column stays NULL.
- **Read path:** `get_current_bookmark_response` emits `DeviceId` and `DeviceModel` only when non-null, so legacy NULL rows don't leak empty keys into the response.

## Compatibility

- **Existing rows:** both columns are NULL until the next write. Response omits the new keys for those rows — old client behavior preserved.
- **Old clients:** ignore unknown response keys (the Kobo JSON parser is tolerant), so adding `DeviceId` / `DeviceModel` doesn't break anything.
- **Headers absent on the request:** captured as `None`, stored as NULL — same as legacy rows.

## Test plan

- [x] Fresh install: columns created on first boot via migration.
- [x] Existing install (with prior `kobo_bookmark` data): migration adds both columns, existing rows surface `NULL` for both.
- [x] Real Kobo Libra Colour write: captured headers were `x-kobo-deviceid: af1fbd73…d752e8` and `x-kobo-devicemodel: Kobo Libra Colour` (verified via temporary header-logging build); both stored correctly.
- [x] GET bookmark for unwritten book: response omits `DeviceId` and `DeviceModel`.
- [x] Custom (non-Kobo-firmware) iOS client that also sends both headers: tracked correctly, distinct from Kobo's identifiers.

## Why both columns

`device_id` alone is sufficient for *identity* but always opaque. `device_model` alone is *readable* but not unique (two devices of the same model collide). Storing both keeps the unique identity for write-source tracking *and* gives the client something to display without needing a separate UUID-to-name mapping.

## Note

One-of-two bookmark fixes split from the same internal branch. The companion PR (#1362) preserves `0.0` progress in the bookmark response — independent fix, reviewable separately.

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -945,6 +945,7 @@ def HandleStateRequest(book_uuid):
                 current_bookmark = kobo_reading_state.current_bookmark
                 current_bookmark.progress_percent = request_bookmark["ProgressPercent"]
                 current_bookmark.content_source_progress_percent = request_bookmark["ContentSourceProgressPercent"]
+                current_bookmark.device_id = request.headers.get('x-kobo-deviceid')
                 location = request_bookmark["Location"]
                 if location:
                     current_bookmark.location_value = location["Value"]
@@ -1106,6 +1107,8 @@ def get_current_bookmark_response(current_bookmark):
             "Type": current_bookmark.location_type,
             "Source": current_bookmark.location_source,
         }
+    if current_bookmark.device_id is not None:
+        resp["DeviceId"] = current_bookmark.device_id
     return resp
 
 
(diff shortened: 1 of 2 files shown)


raphi011:
Since this is not part of kobo's official sync protocol we may not want to add any custom fields just for convenience.

Issue #1325, opened 2026-05-02
[Bug? Request?] Respect subtitle column (Kobo)


szkafen:
There's a configuration in Calibre Desktop + the Kobo plugins that allows me to set a column as a "subtitle", the contents of which then shows in the Kobo eReader itself (Libra 2 in my case). It can be text or numbers, so some people use it to add the series, the collection, or whatever info is suitable for each.

With Calibre Desktop, I had that column fed with the Count Pages plugin so that, on the Libra, it would show "x Pages" under the title of each book. Furthermore, for books on which I've already done this on Calibre Desktop, CWA does show the number of pages.

I've noticed CWA does not do this. If I sync the Kobo with CWA, it will erase all subtitles (easily fixable just pluging in the Kobo and waiting for Calibre Desktop to update the metadata). It only takes a minute or so in total, but nevertheless it's annoying having to do so every time I sync.

I'm wondering if there's a way for CWA to do this:

1- Once a book is ingested, converted and added, the Count Pages plugin does it's thing and feeds the Subtitle column
2- The page number then appears in CWA
3- Once sync with the Kobo is done, the subtitle appears as intended

AsherMaximum:
I added this in PR #1375 for subtitle, and #1374 for page & word count (and ISBN as well)

The configuration for the displayed subtitle in CWA and what is sent to the Kobo is different, so if you want to send the page count as the subtitle, so it displays underneath the book in the library view, you can. In that case you only need #1375. Or you can send pages as the actual pages field to the Kobo, in which case it will show up under the Details section of the book. That part of the change comes from #1374

PR #1375, opened 2026-05-27
Kobo Subtitle configuration option

AsherMaximum:
Copy of changes from https://github.com/janeczku/calibre-web/pull/3376

Issue #1325


szkafen:
Just wanted to add my support for this PR, it's exactly the solution to the issue I reported in #1325. Being able to configure a subtitle column for Kobo sync would be a great QoL improvement.

PR #1406, opened 2026-06-21
fix(mail): resolve TypeError mismatch during eReader task execution

kurtlieber:
Resolves an unhandled exception inside the background task runner where ad-hoc or alternative recipient routing configurations passed structured dictionaries into string compilation and join blocks instead of raw text strings.  This would cause some "Send to eReader" attempts to fail with a "sequence item 0: expected str instance, dict found"
SCR-20260621-cmzp

### DIFF
--- a/cps/tasks/mail.py
+++ b/cps/tasks/mail.py
@@ -167,7 +167,22 @@ class TaskEmail(CalibreTask):
             elif hasattr(e, "message"):
                 text = e.message
             elif hasattr(e, "args"):
-                text = '\n'.join(e.args)
+                if isinstance(e, smtplib.SMTPRecipientsRefused):
+                    refused = []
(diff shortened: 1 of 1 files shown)


new-usemame:
Fixed in [Calibre-Web-NextGen](https://github.com/new-usemame/Calibre-Web-NextGen) v4.0.171 ([`05579e16`](https://github.com/new-usemame/Calibre-Web-NextGen/commit/05579e16b)) — an independent take on the same bug, with thanks to @kurtlieber for pinning it down.

If your "Send to eReader" dies with a `TypeError` instead of the real mail-server reason, the community-maintained CWA build now shows the actual rejection (e.g. `550 User unknown`):

```
docker pull ghcr.io/new-usemame/calibre-web-nextgen:latest
```

PR #1435, opened 2026-07-16
Fix duplicate auto-resolution cooldown: UTC/local timestamp mismatch + settings bool-coercion#1

### LINKED ISSUE #1434: Duplicate auto-resolution cooldown broken by UTC/local timestamp mismatch (+ settings bool-coercion bug)
### Summary

The duplicate-detection auto-resolution cooldown (`duplicate_auto_resolve_cooldown_minutes`) does not honor its configured value. In practice it either collapses to ~1 minute (masking the bug most of the time) or, when two resolutions land close together, appears to get "stuck" reporting a cooldown of roughly **420 minutes (7 hours) too much**, decaying in real time as if a phantom resolution had just happened. The 420-minute figure is not a hardcoded constant — it is the host's UTC offset (`America/Los_Angeles` = UTC−7 in DST) leaking into a naive-datetime comparison. This is a genuine bug, not a timezone misconfiguration on the deployer's side: it reproduces with a correctly-configured `TZ` env var, and no `TZ` setting can fix it (see Root Cause).

Confirmed present, byte-for-byte identical, on `crocodilestick/Calibre-Web-Automated@main` (commit `6f5f5f3e3818f34c3e8bae3f83fda6e5613157ed`) as of 2026-07-15. Also present in the `ghcr.io/new-usemame/calibre-web-nextgen:v4.1.10` fork, inherited unchanged — this is not fork-introduced.

### Environment

- Upstream repo: `crocodilestick/Calibre-Web-Automated`, `main` @ `6f5f5f3e3818f34c3e8bae3f83fda6e5613157ed`
- Also reproduced on: `ghcr.io/new-usemame/calibre-web-nextgen:v4.1.10`
- Container `TZ=America/Los_Angeles` (correctly set in `docker-compose.yml`)
- SQLite (bundled), Python 3.13

### Symptom

With `duplicate_auto_resolve_enabled=1` and `duplicate_auto_resolve_cooldown_minutes` set to a small nonzero value (e.g. `5`), auto-resolution intermittently refuses to run, logging:

```
[cwa-duplicates] Auto-resolution skipped due to cooldown: 192.0 minutes remaining
```

...even though the configured cooldown is 5 minutes and the last successful auto-resolution was under 4 hours ago. The remaining-minutes value decays in real time (192.0 → 189.4 a few minutes later), consistent with a fixed but wrong baseline rather than a static hardcoded number.

### Root cause

Two independent bugs compound to produce this symptom.

#### Bug 1 — `duplicate_auto_resolve_cooldown_minutes` is missing from the `integer_settings` allowlist

In `scripts/cwa_db.py`, `CWA_DB.get_cwa_settings()` converts any settings-table column holding a plain SQLite `INTEGER` value to a Python `bool`, *unless* the column name is explicitly listed in `integer_settings`:

https://github.com/crocodilestick/Calibre-Web-Automated/blob/6f5f5f3e3818f34c3e8bae3f83fda6e5613157ed/scripts/cwa_db.py#L523

```python
integer_settings = ['ingest_timeout_minutes', 'ingest_stale_temp_minutes', 'ingest_stale_temp_interval', 'auto_send_delay_minutes', 'hardcover_auto_fetch_batch_size', 'hardcover_auto_fetch_schedule_hour', 'duplicate_scan_hour', 'duplicate_scan_chunk_size', 'duplicate_scan_debounce_seconds', 'archived_cleanup_schedule_hour', 'cover_download_max_mb']
```

`duplicate_auto_resolve_cooldown_minutes` is absent from this list, so any nonzero value (e.g. `5`) gets bool-coerced: `bool(5) == True`. `cps/tasks/duplicate_scan.py` then does:

```python
cooldown_minutes = int(cwa_db.cwa_settings.get('duplicate_auto_resolve_cooldown_minutes', 0))
```

`int(True)` is `1` — so the *effective* cooldown is always exactly 1 minute regardless of what the admin configures, or `0` (disabled) only when the stored value happens to be `0` itself.

Note there is already a second, independently-maintained copy of this same allowlist logic in the settings-page route handler (`set_cwa_settings()`, `cps/cwa_functions.py`) that **does** include `duplicate_auto_resolve_cooldown_minutes`. The two lists have drifted apart — the one actually used for enforcement (`cwa_db.py`) is the broken one.

**Confirmed live** via:
```python
from cwa_db import CWA_DB
db = CWA_DB()
db.cwa_settings.get('duplicate_auto_resolve_cooldown_minutes')  # -> True (bool), not 5
```

#### Bug 2 — `cwa_duplicate_resolutions.timestamp` is written in UTC but compared against local time

Schema:

https://github.com/crocodilestick/Calibre-Web-Automated/blob/6f5f5f3e3818f34c3e8bae3f83fda6e5613157ed/scripts/cwa_schema.sql#L200-L213

```sql
CREATE TABLE IF NOT EXISTS cwa_duplicate_resolutions (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    timestamp DATETIME DEFAULT CURRENT_TIMESTAMP,
    ...
```

`log_duplicate_resolution()` never overrides this default with an explicit value:

https://github.com/crocodilestick/Calibre-Web-Automated/blob/6f5f5f3e3818f34c3e8bae3f83fda6e5613157ed/scripts/cwa_db.py#L2493-L2509

```python
def log_duplicate_resolution(self, group_hash, group_title, group_author, kept_book_id,
                             deleted_book_ids, strategy, trigger_type, user_id=None, notes=None):
    self.cur.execute("""
        INSERT INTO cwa_duplicate_resolutions
        (group_hash, group_title, group_author, kept_book_id, deleted_book_ids,
         strategy, trigger_type, user_id, notes)
        VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)
    """, (...))
```

SQLite's `CURRENT_TIMESTAMP` is **always UTC**, per the SQL standard — independent of the host/container `TZ` setting. Meanwhile, the cooldown check in `cps/tasks/duplicate_scan.py` reads that value back and compares it against Python's *local* time:

https://github.com/crocodilestick/Calibre-Web-Automated/blob/6f5f5f3e3818f34c3e8bae3f83fda6e5613157ed/cps/tasks/duplicate_scan.py#L253-L266

```python
last_resolution = cwa_db.cur.execute("""
    SELECT MAX(timestamp) FROM cwa_duplicate_resolutions
    WHERE trigger_type='automatic'
""").fetchone()[0]

if last_resolution:
    last_time = datetime.fromisoformat(last_resolution)
    now = datetime.now()
    elapsed = (now - last_time).total_seconds() / 60

    if elapsed < cooldown_minutes:
        remaining = cooldown_minutes - elapsed
        log.info("[cwa-duplicates] Auto-resolution skipped due to cooldown: %.1f minutes remaining", remaining)
        return
```

Both `last_time` and `now` are naive `datetime` objects with no timezone attached, so Python compares them literally. A resolution that really happened at 23:11:45 local (PDT, UTC−7) gets stored as `06:11:45`, which — compared against a later local `datetime.now()` — reads as roughly 7 hours in the *future*. That makes `elapsed` strongly negative, which inflates `remaining` by close to 420 minutes (7h × 60). This is the entire source of the "~420 minutes" figure; it is not a hardcoded default anywhere in the code.

This is silent under normal conditions because MAX(timestamp) still generally orders correctly and gaps between auto-resolutions are usually well over 7 hours + cooldown, so the bug never surfaces. It only manifests when two duplicate groups resolve close together (in our repro, ~34 seconds apart) and the second resolution's cooldown check lands within the artificial ~7-hour phantom window.

Every other stats table in the codebase (`cwa_import`, `cwa_conversions`, `epub_fixes`, and even `cwa_duplicate_cache.scan_timestamp` in this same file) explicitly passes `datetime.now().isoformat()` / `strftime(...)` at insert time rather than relying on the schema's `CURRENT_TIMESTAMP` default — `cwa_duplicate_resolutions` is the outlier that doesn't follow this convention.

(Also worth a look, same anti-pattern but not confirmed to be live-compared against local time yet: `cwa_duplicate_book_keys.updated_at` in `cps/duplicate_index.py` is written via literal SQL `CURRENT_TIMESTAMP` too. It's currently write-only in the code we checked, so not an active bug, but it's the same footgun and would bite the moment anything compares it against `datetime.now()`.)

### Reproduction

1. Set `duplicate_auto_resolve_enabled=1`, `duplicate_auto_resolve_cooldown_minutes=5`, in a non-UTC timezone with a positive UTC offset in effect (e.g. any US timezone).
2. Trigger two auto-resolutions of different duplicate groups within roughly `(cooldown_minutes + TZ_offset_minutes)` of each other — e.g. under 7h05m apart in `America/Los_Angeles` (PDT).
3. The second resolution's cooldown check compares its local `datetime.now()` against the first resolution's UTC-stamped `timestamp`, computes a large negative `elapsed`, and skips with a `remaining` value inflated by ~`TZ_offset_minutes`.

Minimal repro of the arithmetic, no app needed:

```python
from datetime import datetime
last = datetime.fromisoformat('2026-07-15 06:11:45')   # stored as UTC via CURRENT_TIMESTAMP
now  = datetime.fromisoformat('2026-07-15 03:03:22')   # actual local time when check runs
elapsed = (now - last).total_seconds() / 60
print(elapsed)                    # -188.38  (negative!)
print(1 - elapsed)                # 189.38  ("minutes remaining", matches observed log line)
```

### Suggested fix

1. **`scripts/cwa_db.py`, `get_cwa_settings()`** — add `duplicate_auto_resolve_cooldown_minutes` to the `integer_settings` list (matching the copy already in `cps/cwa_functions.py`'s `set_cwa_settings()`), so it stops being bool-coerced.

2. **`scripts/cwa_db.py`, `log_duplicate_resolution()`** — pass an explicit local timestamp instead of relying on the schema default, matching every other stats-table writer in this file:

```python
timestamp = datetime.now().strftime('%Y-%m-%d %H:%M:%S')
self.cur.execute("""
    INSERT INTO cwa_duplicate_resolutions
    (timestamp, group_hash, group_title, group_author, kept_book_id, deleted_book_ids,
     strategy, trigger_type, user_id, notes)
    VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
""", (timestamp, group_hash, group_title, group_author, kept_book_id,
      json.dumps(deleted_book_ids), strategy, trigger_type, user_id, notes))
```

   (Longer-term, `cwa_duplicate_resolutions.timestamp DATETIME DEFAULT CURRENT_TIMESTAMP` in `cwa_schema.sql` is a footgun regardless — worth dropping the `DEFAULT` or switching every reader to `datetime('now')`/UTC-aware comparisons consistently, so a future writer can't reintroduce this by omitting the explicit value again.)

3. Separately (not on the hot path here, but same root defect and worth a one-line note): `update_cwa_settings()` in `cwa_db.py` writes new setting values to the DB but never refreshes `self.cwa_settings` afterward. The one caller found (`cps/cwa_functions.py`'s settings-page route) already works around this by manually calling `cwa_settings = cwa_db.get_cwa_settings()` right after, so it isn't currently causing visible symptoms — but any future caller that doesn't do that workaround will silently read stale settings for the lifetime of that `CWA_DB` instance.

Happy to open a PR with these three changes if that's useful — I already have a tested patch (verified end-to-end against a real duplicate group, confirmed the cooldown check and auto-resolution both behave correctly post-fix, and confirmed against upstream `main` that the bug is present there too, not fork-introduced).

jdbway:
Adds duplicate_auto_resolve_cooldown_minutes to the integer_settings allowlist in get_cwa_settings() so it stops being bool-coerced (any nonzero value was collapsing to an effective 1-minute cooldown).
    log_duplicate_resolution() now stamps an explicit local timestamp instead of relying on the schema's DEFAULT CURRENT_TIMESTAMP (which SQLite always stores as UTC), matching every other stats-table writer in this file. This was the actual source of the "~420 minutes remaining" symptom described in https://github.com/crocodilestick/Calibre-Web-Automated/issues/1434 — a naive UTC-vs-local comparison in the cooldown check.
    Refreshes self.cwa_settings at the end of update_cwa_settings() so callers that don't manually re-fetch don't silently read stale values.

Fixes https://github.com/crocodilestick/Calibre-Web-Automated/issues/1434

### DIFF
--- a/scripts/cwa_db.py
+++ b/scripts/cwa_db.py
@@ -520,7 +520,7 @@ class CWA_DB:
                 cwa_settings[key] = default_value
 
         # Define which settings should remain as integers (not converted to boolean)
-        integer_settings = ['ingest_timeout_minutes', 'ingest_stale_temp_minutes', 'ingest_stale_temp_interval', 'auto_send_delay_minutes', 'hardcover_auto_fetch_batch_size', 'hardcover_auto_fetch_schedule_hour', 'duplicate_scan_hour', 'duplicate_scan_chunk_size', 'duplicate_scan_debounce_seconds', 'archived_cleanup_schedule_hour', 'cover_download_max_mb']
+        integer_settings = ['ingest_timeout_minutes', 'ingest_stale_temp_minutes', 'ingest_stale_temp_interval', 'auto_send_delay_minutes', 'hardcover_auto_fetch_batch_size', 'hardcover_auto_fetch_schedule_hour', 'duplicate_scan_hour', 'duplicate_scan_chunk_size', 'duplicate_scan_debounce_seconds', 'duplicate_auto_resolve_cooldown_minutes', 'archived_cleanup_schedule_hour', 'cover_download_max_mb']
         
         # Define which settings should remain as floats (not converted to boolean)
         float_settings = ['hardcover_auto_fetch_min_confidence', 'hardcover_auto_fetch_rate_limit']
@@ -556,6 +556,7 @@ class CWA_DB:
                 # Continue to next setting instead of failing completely
                 continue
         self.set_default_settings()
+        self.cwa_settings = self.get_cwa_settings()
 
 
     def enforce_add_entry_from_log(self, log_info: dict, trigger_type: str = "auto -log"):
@@ -2490,17 +2491,22 @@ class CWA_DB:
             print(f"[cwa-db] Error updating duplicate cache: {e}")
             return False
 
-    def log_duplicate_resolution(self, group_hash, group_title, group_author, kept_book_id, 
+    def log_duplicate_resolution(self, group_hash, group_title, group_author, kept_book_id,
                                  deleted_book_ids, strategy, trigger_type, user_id=None, notes=None):
         """Log a duplicate resolution to audit table"""
         import json
         try:
+            # Explicit local timestamp: the column's schema DEFAULT CURRENT_TIMESTAMP is
+            # UTC, but duplicate_scan.py's cooldown check compares against datetime.now()
+            # (local). Passing it explicitly keeps this table consistent with every other
+            # stats table (cwa_import, cwa_conversions, etc.), which all stamp local time.
+            timestamp = datetime.now().strftime('%Y-%m-%d %H:%M:%S')
             self.cur.execute("""
-                INSERT INTO cwa_duplicate_resolutions 
-                (group_hash, group_title, group_author, kept_book_id, deleted_book_ids, 
+                INSERT INTO cwa_duplicate_resolutions
+                (timestamp, group_hash, group_title, group_author, kept_book_id, deleted_book_ids,
                  strategy, trigger_type, user_id, notes)
-                VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)
-            """, (group_hash, group_title, group_author, kept_book_id, 
+                VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+            """, (timestamp, group_hash, group_title, group_author, kept_book_id,
                   json.dumps(deleted_book_ids), strategy, trigger_type, user_id, notes))
             self.con.commit()
             return True


jdbway:
Heads up on two things that came out of a related fix in another codebase.
I also filed this same bug against new-usemame/calibre-web-nextgen (the fork I actually run), since it isn't a tracked GitHub fork of this repo and wouldn't get this fix automatically. The maintainer there built on this diagnosis but shipped a slightly different (and more thorough) fix — worth passing along before this merges:
Testing gotcha: if anyone adds a regression test for this, it needs to force a non-UTC TZ. If CI runs in UTC (a common default), local time and UTC time become identical, so this bug becomes untestable — a test could pass even against the broken code, since there's no timezone offset left to expose the mismatch.
Read-side vs. write-side: this PR fixes the timestamp bug by making the write (log_duplicate_resolution) stamp local time instead of relying on the UTC schema default. That's correct going forward, but rows already sitting in cwa_duplicate_resolutions from before this fix are still UTC-stamped — so SELECT MAX(timestamp) over a mix of old-UTC and new-local rows can still return a stale UTC row and mis-trigger the cooldown on existing installs. Their fix instead makes the read side (the cooldown comparison in duplicate_scan.py) compare in UTC, leaving the write side untouched — that handles old and new rows the same way with no migration needed, since rows already on disk are UTC.
Not asking anyone to change this PR — just flagging both in case they're useful before merge. Their fix, for reference: new-usemame/calibre-web-nextgen#950

PR #1437, opened 2026-07-19
Allow kobo without account to sync correctly

theraspb3rry:
Finally got around to playing with a freshly factory reset Kobo and this patch (which seems to have been scrubbed from the original [owner](https://github.com/mihneabulu)).

I needed the `/v1/user/add-device` route to return 200 on a POST so that's done there and now my fresh device setup following the article below successfully syncs the first time :)
https://jacobalbano.com/2021/06/24/setting-up-a-kobo-without-logging-in/

Happy to make any other required changes, cheers.

### DIFF
--- a/cps/kobo.py
+++ b/cps/kobo.py
@@ -376,7 +376,7 @@ def HandleSyncRequest():
             .all()
 
         new_tags_last_modified = sync_token.tags_last_modified
-            
+
         for shelf in magic_shelves:
             books, _ = magic_shelf.get_books_for_magic_shelf(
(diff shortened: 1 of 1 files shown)


crocodilestick:
Superseded by #1367 which provides a more comprehensive fix for Kobo fully-local sync on modern firmware — includes both OIDC discovery AND the reading-services stub (before_app_request hook) that catches Kobo API calls and prevents sync aborts. Thanks for the contribution!

theraspb3rry:
Thank you so much!

PR #1474, opened 2026-08-07
Fix split library paths in CWA workflows

iuyo5678:
Summary

- generate post-ingest KOReader checksums against the configured `metadata.db`, not the split book-files root
- back up automatically resolved duplicate books from `config.get_book_path()`
- run `calibredb check_library` and `restore_database` with the book-files root plus `CALIBRE_OVERRIDE_DATABASE_PATH` in split-library mode
- preserve the existing behavior for non-split libraries

## Problem

When **Separate Book Files from Library** is enabled, CWA keeps `metadata.db` in `config_calibre_dir` and stores book files in `config_calibre_split_dir`. A few workflows still treated `config_calibre_dir` and the book-files root as the same directory. This caused checksum generation to open a non-existent database under the books directory, duplicate resolution to look for backup sources under the database directory, and Calibre database check/restore commands to receive the wrong library context.

## Tests

```text
pytest -q -m unit \
  tests/unit/test_calibre_cli.py \
  tests/unit/test_ingest_batch_dirty.py \
  tests/unit/test_duplicate_delete_index_maintenance.py

7 passed, 4 deselected
```

Additional checks:

- `git diff --check`
- Python `compileall` for all changed Python files
- Ruff `E4,E7,E9,F` checks for the new helper and its tests

### DIFF
--- a/cps/admin.py
+++ b/cps/admin.py
@@ -38,6 +38,7 @@ from . import db, calibre_db, ub, web_server, config, updater_thread, gdriveutil
 from .helper import check_valid_domain, send_test_mail, reset_password, generate_password_hash, check_email, \
     valid_email, check_username
 from .embed_helper import get_calibre_binarypath
+from .calibre_cli import get_calibre_cli_context
 from .gdriveutils import is_gdrive_ready, gdrive_support
 from .render_template import render_title_template, get_sidebar_config
 from .services.worker import WorkerThread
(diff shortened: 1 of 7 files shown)


iuyo5678:
I pushed a follow-up commit that restores the stubbed modules after the new
unit test, preventing pytest worker state from leaking into subsequent tests.

The original failing tests now pass locally in the same xdist worker.
Could a maintainer please approve and run the updated workflow?

